PR #174, opened 2017-02-03
Add a Text Label to the Mute Image Button

RachelOlivero:
Adds a screen reader accessible label to the mute button. Label switches between mute and unmute to match the action that will be taken when the button is next pressed.


DSheirer:
Thanks Rachel!

PR #232, opened 2017-03-23
Resolves #229 - duplicate P25 TDMA call events.

### LINKED ISSUE #229: P25 Phase 2 Call Events are duplicated
A P25 Phase 2 (TDMA) call event is generated each time the control channel references an in-progress call.  Update the P25ChannelState to track Phase 2 calls, even though they are not being monitored for audio output.

DSheirer:
Resolves #229


bullwinkle041:
I seem to have the same problem b25 phase one is there any apps that can fix this problem

Issue #250, opened 2017-06-26
ChannelState's StateMonitor should catch Throwable errors


DSheirer:
Update the ChannelState StateMonitor runnable to catch Throwable errors and log them.  

Users are reporting that channels sometimes get stuck in the TEARDOWN state and never progress/return to the IDLE state, effectively locking out that frequency until the application is restarted.  This may be an indication that the StateMonitor runnable has died ... possible from a Throwable error.

PR #251, opened 2017-07-01
StateMonitor - try/catch for Throwables

DSheirer:
Updates the StateMonitor runnable to catch Throwables while monitoring to set state to IDLE.


DSheirer:
Resolves #250

PR #301, opened 2017-12-16
SDRTrunk upgrade

tsypuk:
Hy Denny, my name is Roman, I'm Radio HAM and software developer.
I was looking for project, modules that are using SDR from java and have found out your SDRTrunk.
Great project thanks a lot for your work.

By this change set I propose you the following:
1. Added gradle build system (most robust nowadays :) )
2. Most part of dependencies are removed from local folder and are under control build system.
3. Dependencies versions are upgraded to latest.
4. Project structure changed based to java project.
5. Configuration and property files are moved to resources folder.
6. Created uber-jar artifact with incremental versioning. All dependencies and project are build into single jar file (no need in class path, easy to execute, copy, build, run).
7. Updated README.md

Notes:
I have moved to ua.in.smartjava project based on domain design but if you have your own domain name feel free to change it. I had no find yet external dependecies for this libraries
- java-lame-3.98.4.jar
- jmbe-interface-0.3.3.jar
- radioreference-14.jar
- wsdl4j-1.5.1.jar
so I left them in imports folder, all other are in build system.
To changed java project structure you can add test, based on JUnit for example, to guarantee that your code works correct.

I have build project on my system and successfully run it with RT820T. But in my country there is no trunking network mostly radio channel. So please test that it works correct.


DSheirer:
Hi Roman!  

Thanks for the pull request!  I like the gradle build system.

I updated the package structure to use 'io.github.dsheirer'.  Do you want to eventually push releases to maven central or jcenter?

I'm working on resolving a few issues and then I'll start testing and hope to merge this soon.  

cheers,
Denny

tsypuk:
Hi Denny, great news !

Currently gradle project is configured to build only local jar file. 
Personally, I think it would be great to create different layers of abstraction and extract for example all code that interacts only with radio module into sdr-radio-library. SdrTrunk GUI would import this radio-library and interact with it. If so in future this radio-library can be versioned and pushed to maven or jcenter. 

This would allow other users to use radio-library in their project (for web-ui, CLI or socket server for ex.). And would make more safe adding GUI features or fixes.

Issue #353, opened 2018-05-06
Forward Slash in site or channel name causes recording error


DSheirer:
I found a problem that caused recording to stop for certain channels, and caused these errors:
20180504 142559.576 [sdrtrunk pool-1-thread-4] ERROR i.g.d.record.RecorderManager - Error attempting to start new audio wave recorder. All (future) audio recording is disabled  [110MB/603MB 18%]

java.nio.file.InvalidPathException: Trailing char < > at index 50: C:\Users\xxxxxxx \SDRTrunk\recordings\xxxxxx_50C1 / 20673_TRAFFIC__TO_000-7216_FROM_000-0467_20180504_142559.576.tmp

The problem was a forward slash in the site names. Once I replaced it with underscore and restarted I was able to record those channels again.

DSheirer:
@whelgeson Thank you very much for resolving this issue!

PR #448, opened 2018-12-15
Fixing issue #353 Forward Slash in site or channel name causes recording error

whelgeson:
Moved replaceIllegalCharacters method into StringUtils.
Modified classes that write out to disk to call this method when saving a file.

### DIFF
--- a/src/main/java/io/github/dsheirer/module/log/EventLogger.java
+++ b/src/main/java/io/github/dsheirer/module/log/EventLogger.java
@@ -19,7 +19,6 @@
 package io.github.dsheirer.module.log;
 
 import io.github.dsheirer.module.Module;
-import io.github.dsheirer.util.StringUtils;
 import io.github.dsheirer.util.TimeStamp;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -73,7 +72,7 @@ public abstract class EventLogger extends Module
                 sb.append(File.separator);
                 sb.append(TimeStamp.getTimeStamp("_"));
                 sb.append("_");
-                sb.append(StringUtils.replaceIllegalCharacters(mFileNameSuffix));
+                sb.append(mFileNameSuffix);
 
                 mLogFileName = sb.toString();
 
--- a/src/main/java/io/github/dsheirer/module/log/EventLogger.java
+++ b/src/main/java/io/github/dsheirer/module/log/EventLogger.java
@@ -19,6 +19,7 @@
 package io.github.dsheirer.module.log;
 
 import io.github.dsheirer.module.Module;
+import io.github.dsheirer.util.StringUtils;
 import io.github.dsheirer.util.TimeStamp;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -34,9 +35,6 @@ import java.nio.file.Path;
 public abstract class EventLogger extends Module
 {
     private final static Logger mLog = LoggerFactory.getLogger(EventLogger.class);
-
-    private static final String[] ILLEGAL_FILENAME_CHARACTERS = {"#", "%", "&", "{", "}", "\\", "<", ">",
-        "*", "?", "/", " ", "$", "!", "'", "\"", ":", "@", "+", "`", "|", "="};
 
     private Path mLogDirectory;
     private String mFileNameSuffix;
@@ -75,7 +73,7 @@ public abstract class EventLogger extends Module
                 sb.append(File.separator);
                 sb.append(TimeStamp.getTimeStamp("_"));
                 sb.append("_");
-                sb.append(replaceIllegalCharacters(mFileNameSuffix));
+                sb.append(StringUtils.replaceIllegalCharacters(mFileNameSuffix));
 
                 mLogFileName = sb.toString();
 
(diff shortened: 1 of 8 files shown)


src/main/java/io/github/dsheirer/module/log/EventLogger.java
```diff
@@ -75,7 +73,7 @@ public void start()
                 sb.append(File.separator);
                 sb.append(TimeStamp.getTimeStamp("_"));
                 sb.append("_");
-                sb.append(replaceIllegalCharacters(mFileNameSuffix));
+                sb.append(StringUtils.replaceIllegalCharacters(mFileNameSuffix));
```
DSheirer: Can you please move the cleaning method up to the ChannelProcessingManager:290 and clean the channel.getName() value.  The current approach causes path separators in linux to be converted to underscores and causes an illegal filepath.


        /* Setup event logging */
        List<Module> loggers = mEventLogManager.getLoggers(channel.getEventLogConfiguration(),
            StringUtils.replaceIllegalCharacters(channel.getName()));

src/main/java/io/github/dsheirer/record/RecorderManager.java
```diff
@@ -225,7 +226,7 @@ private String getFilePrefix(ReusableAudioPacket packet)
             }
         }
 
-        return sb.toString();
+        return StringUtils.replaceIllegalCharacters(sb.toString());
```
DSheirer: Could you please move the clean method to where each of the System/Site/Name values are added.  Otherwise, the clean method causes invalid linux filepaths.  

        sb.append(metadata.hasChannelConfigurationSystem() ? StringUtils.replaceIllegalCharacters(metadata.getChannelConfigurationSystem()) + "_" : "");
        sb.append(metadata.hasChannelConfigurationSite() ? StringUtils.replaceIllegalCharacters(metadata.getChannelConfigurationSite()) + "_" : "");
        sb.append(metadata.hasChannelConfigurationName() ? StringUtils.replaceIllegalCharacters(metadata.getChannelConfigurationName()) + "_" : "");

src/main/java/io/github/dsheirer/record/wave/AudioPacketWaveRecorder.java
```diff
@@ -110,7 +111,7 @@ public void start()
                 sb.append(TimeStamp.getLongTimeStamp("_"));
                 sb.append(".tmp");
 
-                mFile = Paths.get(sb.toString());
+                mFile = Paths.get(StringUtils.replaceIllegalCharacters(sb.toString()));
```
DSheirer: This can be removed with the changes from review comment #2 so that linux file paths won't get corrupted.

src/main/java/io/github/dsheirer/record/wave/ComplexBufferWaveRecorder.java
```diff
@@ -89,8 +90,8 @@ public void start()
                 sb.append("_");
                 sb.append(TimeStamp.getTimeStamp("_"));
                 sb.append(".wav");
-
-                mFile = Paths.get(sb.toString());
+                String cleaned = StringUtils.replaceIllegalCharacters(sb.toString());
```
DSheirer: Could you please move this clean method up into the RecorderManager.getBasebandRecorder() method so that linux filepaths don't get corrupted:

For example:

        StringBuilder sb = new StringBuilder();
        sb.append(getRecordingBasePath());
        sb.append(File.separator).append(StringUtils.replaceIllegalCharacters(channelName)).append("_baseband");
        return new ComplexBufferWaveRecorder(BASEBAND_SAMPLE_RATE, sb.toString());

src/main/java/io/github/dsheirer/record/wave/WaveWriter.java
```diff
@@ -155,7 +156,8 @@ private void rename() throws IOException
             {
                 String renamedWithSuffix = mFile.getFileName().toString();
                 renamedWithSuffix = renamedWithSuffix.replace(".tmp", "_" + suffix + ".wav");
-                Path renamedWithSuffixPath = mFile.resolveSibling(renamedWithSuffix);
+                String cleaned = StringUtils.replaceIllegalCharacters(renamedWithSuffix);
```
DSheirer: This change can be removed.  The original filename should have valid characters if it got to this point, and so the updated suffix version should also have valid filename charactes.  Otherwise, this will corrupt linux filepaths.

src/main/java/io/github/dsheirer/util/StringUtils.java
```diff
@@ -20,6 +20,9 @@
 
 public class StringUtils
```
DSheirer: This is a great addition and makes it easy to apply across the codebase.  Thanks!!

src/main/java/io/github/dsheirer/record/RecorderManager.java
whelgeson: I changed this suggestion a little bit to encompass the 'metadata.getPrimaryAddressTo().getIdentifier()'  calls as well. My thinking is that it may somehow be possible to have an illegal char in the identifier.

PR #470, opened 2018-12-28
467 Alias List Update On Alias Delete/Change

DSheirer:
Updates alias lists to fully clean aliases from the alias list when deleted and to perform a delete and re-add on any alias that changes, to ensure that the alias list is kept fully up to date.

Removed site id as an alias identifier.

Added support for converting ltr-net uid's to talkgroups in playlist
updater and updated playlist to version 3 to ensure anyone using current
codebase will have their playlists fully updated.

Moved MIN, UniqueID and Site ID to the legacy package and updated
label descriptions to signify that they are no longer supported.


DSheirer:
This needs to be fully tested/reviewed

PR #617, opened 2019-12-15
#616 Update Build Version to 0.4.0 beta 2.

DSheirer:
Updates radio reference library to 15.1.2 to resolve issue with logback config collision.
Updates OS build names to better reflect processor types: x86_64 (64-bit) or x86_32 (32-bit)


DSheirer:
Resolves #616

PR #658, opened 2020-02-16
#650 Icon Manager User-Added Icons on Windows

### LINKED ISSUE #650: IconManager Can't Add Icons
Reported by user: I can add them using a MAC. But on a Windows machine,.......        it lets me ADD, name the file, brows and select the png file (although it shows greyed out and subdued) and hit enter. It creates a place for it in the icons list, calls it what I named it, but there is no icon there!!
This only happens on Windows, not on a MAC.

DSheirer:
Resolves issue where user added icons were not displaying on windows OS.

Resolves #650

### DIFF
--- a/src/main/java/io/github/dsheirer/icon/Icon.java
+++ b/src/main/java/io/github/dsheirer/icon/Icon.java
@@ -81,7 +81,7 @@ public class Icon implements Comparable<Icon>
         {
             try
             {
-                if(mPath.startsWith("/"))
+                if(!mPath.startsWith("images"))
                 {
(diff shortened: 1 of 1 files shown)


RetiredDude:
How do I add this? Do I have to download and re-compile SDRtrunk?

Thanks,
Jim (one of the two PG county brothers)

hans109h:
Yes, you need to download a clone of the master branch and compile it.  If you have previously used git then all you need to do is "git pull' then "gradlew clean build" or there about...

PR #659, opened 2020-02-17
alias in event log.diff

ImagoTrigger:
puts TO talkgroup alias names in the event logs


DSheirer:
@ImagoTrigger nice, clean implementation.  I like it!  Thanks for the PR.

ImagoTrigger:
nice i use it on my status page

Issue #310, opened 2018-01-01
Update gradle and publish remaining dependencies to maven


DSheirer:
There are 4 library dependencies remaining in the /imports folder following the conversion to the gradle build system.

Push these libraries to an sdrtrunk registered maven repo and update the gradle build script.

PR #664, opened 2020-02-21
Issue #310 - Implementing some build script cleanup work.

whelgeson:
- Adding jitpack to import libs directly from the git repos
 - Removing binary imports from repo
 - Fixing Gradle 7 upgrade warnings


DSheirer:
@whelgeson Awesome!  That helps clean things up.  Thanks very much!

PR #725, opened 2020-04-02
Fix HackRF-JawBreaker and HackRF sample conversion

crsawyer:
- JawBreaker was partially defined but would not initialize

- HackRF samples are signed 8-bit. (If handled as unsigned 8-bit, you will notice a high noise floor and signal reflections around DC)


crsawyer:
Justification for the signed 8-bit change:
Before (unsigned):
HackRF-UnSignedByte
After (signed):
HackRF-SignedByte

DSheirer:
Wow, that's dramatic!  Thanks for the PR.  I'll test this weekend and get it merged in.

DSheirer:
@crsawyer You just turned my HackRF from a $300 paper weight into something useful.  Thanks!!  And thanks for fixing my code errors!

PR #746, opened 2020-04-20
Performance improvements

robekl:


DSheirer:
The performance improviement that you made by eliminating the short:float lookup map is dramatic.  I thought I was being more efficient by using a map ... ugh.

I did a quick comparison between each approach and it was better than 1M:1 speedup with your approach.  Great find!

PR #764, opened 2020-04-27
Use jvm and os

robekl:


DSheirer:
@robekl I'm sorry for not getting to this sooner.  This is a big change set and I wanted to make sure I had some time to spend on profiling, so I had to clear some things off my plate first. I was only able to run the profiler for 15 minutes this morning, but I hope to spend a few hours this weekend. 

I'm attaching two screenshots of the Visual VM profiler against your PR and Master branch.  

The reason I went with the current object reuse approach for buffers is because earlier in the project there was a problem with excessive GC and users with lesser powered CPUs were not keeping up.  I'm intrigued by the heap usage profile that your changes provide and I want to look into it some more.  I share the same philosophy, let the GC handle it.  But at the time, object reuse for the float buffers seemed to be a good solution.

There might be an issue with Thread creation in your PR.  The live thread count continues to grow, although I might not have run it long enough to allow it to plateau.

Profiler against  your PR #764:
profile_pr_764

Profiler against current master branch:
profile_master

DSheirer:
Closing.  Changes incorporated via PR #1148

PR #823, opened 2020-05-20
MPT1327: Fix for call timeout setting not being honored

### LINKED ISSUE #820: MPT1327 - Call timeout is effectively ignored
# The problem

It seems that call timeout setting of MPT1327 decoder is effectively ignored.

In my setup I've set the timeout to 120 seconds:

call-timeout-setting

However, for calls longer than 45 seconds, traffic channel decoding stops after about 45 seconds:

mpt1327-45s-timeout

Part of the spectrum marked by red rectangle shows the traffic channel - as you can see there's a conversation still going on.

After some debugging it seems this behavior is caused by this line in `SingleChannelState`:

https://github.com/DSheirer/sdrtrunk/blob/3b357c4d40cdc990b829d63d59d1528c45465cf3/src/main/java/io/github/dsheirer/channel/state/SingleChannelState.java#L118

... which seems to stop forwarding audio for any traffic channel after `DEFAULT_CALL_TIMEOUT_DELAY_SECONDS` (45) seconds.

# Testing this theory

After replacing usage of mentioned constant with `120_000` (120 seconds) I was able to follow a conversation longer than 120 seconds:

mpt1327-timeout-after-fix

Part of the spectrum marked by red rectangle shows the traffic channel - again, the conversation is still going on: both parts of the call are silent at the moment, but `CLEAR` message wasn't sent just yet. Traffic channel decoding stopped after hardcoded timeout.

# Possible solutions

It seems that call timeout setting for MPT should be used in mentioned line. Couple of fixes come to mind:

## 1. Quick but ugly

Just apply the call timeout for MPT in `SingleChannelState`. More or less:

```java
final int fadeTimeoutSeconds;
if (channel.getDecodeConfiguration() instanceof DecodeConfigMPT1327) {
    fadeTimeoutSeconds = ((DecodeConfigMPT1327) channel.getDecodeConfiguration()).getCallTimeout();
} else {
    fadeTimeoutSeconds = DecodeConfiguration.DEFAULT_CALL_TIMEOUT_DELAY_SECONDS;
}
mStateMachine.setFadeTimeoutBuffer(fadeTimeoutSeconds * 1000);
```

## 2. Introduce a trait interface

```java
interface WithCallTimeout {
    int getCallTimeout();
}
```

Make any interested decoder config implement it:

```java
class DecodeConfigMPT1327 extends DecodeConfiguration implements WithCallTimeout {
}
```

Apply call timeout in `SingleChannelState` based on the trait:

```java
final int fadeTimeoutSeconds;
if (channel.getDecodeConfiguration() instanceof WithCallTimeout) {
    fadeTimeoutSeconds = ((WithCallTimeout) channel.getDecodeConfiguration()).getCallTimeout();
} else {
    fadeTimeoutSeconds = DecodeConfiguration.DEFAULT_CALL_TIMEOUT_DELAY_SECONDS;
}
mStateMachine.setFadeTimeoutBuffer(fadeTimeoutSeconds * 1000);
```

## 3. Add call timeout for all protocol decoders

I.e. to `DecoderConfiguration` base class:

```java
public abstract class DecodeConfiguration extends Configuration
{
    @JsonIgnore
    public int getCallTimeout()
    {
        return DEFAULT_CALL_TIMEOUT_DELAY_SECONDS;
    }
}
```

Then override as needed for specific decoder, e.g. MPT:

```java
@Override
@JsonIgnore(false) // A bit ugly, but needed
@JacksonXmlProperty(isAttribute = true, localName = "call_timeout")
public int getCallTimeout()
{
    return mCallTimeout;
}
```

And of course apply in `SingleChannelState`:

```java
mStateMachine.setFadeTimeoutBuffer(channel.getDecodeConfiguration().getCallTimeout() * 1000);
```

Although I'm not sure if saying that every decoder has a "call timeout" is semantically correct as I don't know much about e.g. P25 or Passport.

***

Please let me know if you think one of these solutions is the way and I will happily submit a PR :slightly_smiling_face:

apankowski:
Fixes #820 

Any requests, remarks & pointers highly welcome 🙂

### DIFF
--- a/src/main/java/io/github/dsheirer/module/decode/DecoderFactory.java
+++ b/src/main/java/io/github/dsheirer/module/decode/DecoderFactory.java
@@ -226,14 +226,18 @@ public class DecoderFactory
         ChannelMap channelMap = channelMapModel.getChannelMap(mptConfig.getChannelMapName());
         Sync sync = mptConfig.getSync();
         modules.add(new MPT1327Decoder(sync));
-        modules.add(new AudioModule(aliasList));
+
+        final int timeout = mptConfig.getCallTimeout() * 1000;
+        // Set max segment audio sample length slightly above call timeout to
+        // not create a new segment if the processing chain finishes a bit after
+        // actual call timeout.
+        modules.add(new AudioModule(aliasList, timeout + 5000));
+
         SourceType sourceType = channel.getSourceConfiguration().getSourceType();
         if(sourceType == SourceType.TUNER || sourceType == SourceType.TUNER_MULTIPLE_FREQUENCIES)
         {
             modules.add(new FMDemodulatorModule(FM_CHANNEL_BANDWIDTH, DEMODULATED_AUDIO_SAMPLE_RATE));
         }
-
-        long timeout = mptConfig.getCallTimeout() * 1000;
 
         if(channelType == ChannelType.STANDARD)
         {
--- a/src/main/java/io/github/dsheirer/module/decode/config/WithCallTimeout.java
+++ b/src/main/java/io/github/dsheirer/module/decode/config/WithCallTimeout.java
@@ -1,0 +1,12 @@
+package io.github.dsheirer.module.decode.config;
+
+/**
+ * Trait of decoder configurations that support providing information about call timeout.
+ */
+public interface WithCallTimeout
+{
+    /**
+     * @return call timeout in seconds
+     */
+    int getCallTimeout();
+}
--- a/src/main/java/io/github/dsheirer/module/decode/config/WithCallTimeout.java
+++ b/src/main/java/io/github/dsheirer/module/decode/config/WithCallTimeout.java
@@ -1,3 +1,25 @@
+/*
+ *
+ *  * ******************************************************************************
+ *  * Copyright (C) 2014-2020 Dennis Sheirer
+ *  *
+ *  * This program is free software: you can redistribute it and/or modify
+ *  * it under the terms of the GNU General Public License as published by
+ *  * the Free Software Foundation, either version 3 of the License, or
+ *  * (at your option) any later version.
+ *  *
+ *  * This program is distributed in the hope that it will be useful,
+ *  * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ *  * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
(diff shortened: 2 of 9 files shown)


src/main/java/io/github/dsheirer/module/decode/DecoderFactory.java
```diff
@@ -226,15 +226,19 @@ private static void processMPT1327(ChannelMapModel channelMapModel, Channel chan
         ChannelMap channelMap = channelMapModel.getChannelMap(mptConfig.getChannelMapName());
         Sync sync = mptConfig.getSync();
         modules.add(new MPT1327Decoder(sync));
-        modules.add(new AudioModule(aliasList));
+
+        final int timeout = mptConfig.getCallTimeout() * 1000;
+        // Set max segment audio sample length slightly above call timeout to
+        // not create a new segment if the processing chain finishes a bit after
+        // actual call timeout.
+        modules.add(new AudioModule(aliasList, timeout + 5000));
```
apankowski, line 234: My intent here, based on your input in the issue description, was to not create next segment in case processing chain will be slightly late with stopping traffic channel processing, mostly for the purposes of recording. But I'm not entirely sure that's the right way to go.

src/main/java/io/github/dsheirer/module/decode/config/WithCallTimeout.java
```diff
@@ -0,0 +1,12 @@
...
+
+/**
+ * Trait of decoder configurations that support providing information about call timeout.
+ */
+public interface WithCallTimeout
+{
+    /**
+     * @return call timeout in seconds
+     */
+    int getCallTimeout();
```
apankowski, line 11: or `getCallTimeoutSeconds` to make it explicit?

src/main/java/io/github/dsheirer/module/decode/DecoderFactory.java
DSheirer: I think the 5 second buffer should be good for handling this case.

src/main/java/io/github/dsheirer/module/decode/config/WithCallTimeout.java
DSheirer: The way you have it coded is good.  The javadoc clarifies it.

DSheirer:
@apankowski Thank you very much for the PR

Issue #884, opened 2020-07-12
Saving a property of an alias when the table is sorted by that property causes IndexOutOfBoundsException (and can corrupt the playlist)


NoahWL:
If the table under Playlist Editor>Aliases>Alias is sorted by a property of an alias (e.g. Color, Icon, Listen) and an alias is saved, an IndexOutOfBoundsException will usually be thrown.  Additionally, if the modified alias is saved using the "Save" button the save confirmation dialog will still appear, and subsequent presses of "Yes" will duplicate that property line in the playlist's XML file.

I can reproduce this error 100% of the time by doing the following:
1. Create or open an alias list with multiple aliases
2. Click on the "Listen" column header to sort the table by that column in ascending order
3. Click on any alias and change the "Priority" box to a different value than what it current is
4. Click the "Save" button.  Despite pressing save, a dialog appears warning that the alias has unsaved changes.
5. Click "YES" on the dialog.  The dialog will reappear and continue to reappear so long as "YES" is selected.
6. Click "NO" to exit the loop (you may have to click "NO" multiple times).  The console will fill with some variation of an IndexOutOfBoundsException on the JFX app thread.

Below is one of the most common exceptions I get.  Depending on how you play around with it you can get a few different traces, but all stem from saving an alias in a way that causes the table's underlying SortedList to update itself mid-save.

<details>
<summary>Exception</summary>
<p>

```
Exception in thread "JavaFX Application Thread" java.lang.IndexOutOfBoundsException: [ fromIndex: 0, toIndex: 2, size: 1 ]
        at com.sun.javafx.scene.control.ReadOnlyUnbackedObservableList.subList(ReadOnlyUnbackedObservableList.java:171)
        at javafx.collections.ListChangeListener$Change.getAddedSubList(ListChangeListener.java:243)
        at com.sun.javafx.scene.control.behavior.TableViewBehaviorBase.lambda$new$1(TableViewBehaviorBase.java:92)
        at com.sun.javafx.collections.ListListenerHelper$Generic.fireValueChangedEvent(ListListenerHelper.java:329)
        at com.sun.javafx.collections.ListListenerHelper.fireValueChangedEvent(ListListenerHelper.java:73)
        at javafx.collections.ObservableListBase.fireChange(ObservableListBase.java:233)
        at com.sun.javafx.scene.control.ReadOnlyUnbackedObservableList.callObservers(ReadOnlyUnbackedObservableList.java:110)
[...]

PR #885, opened 2020-07-12
#884 Wrap setAliases() in Platform.runLater()

NoahWL:
Fixes issue #884 (see issue for more info).

AliasItemEditor.save() modifies objects that are observed by the
TableView's underlying SortedList.  Thus, the TableView's selection can
change by calling save(), which means the ListChangeListener can be
fired.

This is a problem for two reasons.
1. The alert dialogue that prompts the user to save a modified alias
will show when the user presses the "save" button, because saving
modifies the selected alias and causes the list to update.

2. setAliases() calls Alert.showAndWait().  From the JavaDoc:
"It must not be called during animation or layout processing."
This statement is likely there because it can cause issues like the one
here: TableView's layout processing (and the list updating) is paused
and the list is further modified (by calling save()).  Thus, when the
processing is resumed the list is in a different and unexpected state.

Calling setAliases() from Platform.runLater() will allow the SortedList,
TableView, and any other involved objects to finish processing changes
before more modifications are made.  Additionally, it allows save(),
when called by pressing the save button, to complete and set the
modifiedProperty to false, so setAliases() does not display an
additional save prompt.


DSheirer:
@NoahWL Thanks again!

PR #944, opened 2020-10-04
Only update ViewByIdentifierEditor's list when the editor is selected

NoahWL:
Editing many aliases with the AliasBulkEditor becomes very slow when modifying more than just a few aliases.  When applying a change to 1000 aliases, the PlaylistEditor freezes for over a minute.  It turns out most of the processing time came down to running the updateList() method in the ViewByIdentifierEditor class, as it was called multiple times per modified alias.  By changing the editor to only update its list when it is selected updateList() is no longer called in the background unnecessarily.  The same bulk edits take seconds instead of minutes.

I can't help but feel like the way I implemented the change is a bit ugly but it's the best I could come up with without making a lot of changes to AliasViewByIdentifierEditor.  I wanted to change how the list worked so it wouldn't have to be rebuilt every time an alias was added or removed but I wasn't sure what future plans there may be for this class so I settled on something less invasive and easy to change in the future.

### DIFF
--- a/src/main/java/io/github/dsheirer/gui/playlist/alias/AliasEditor.java
+++ b/src/main/java/io/github/dsheirer/gui/playlist/alias/AliasEditor.java
@@ -124,7 +124,7 @@ public class AliasEditor extends TabPane
     {
         if(mAliasViewByIdentifierEditor == null)
         {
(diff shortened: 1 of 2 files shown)


DSheirer:
@NoahWL Thank you!

PR #975, opened 2020-10-22
#740 - Adds Encrypted state to the Status column for P25 Phase 1 Voice Calls

### LINKED ISSUE #740: P25 Phase 1 Doesn't Show Encrypted Status
Phase 1 encrypted audio is being handled correctly, but it doesn't appear that the decoder state is being updated to encrypted and therefore not displaying correctly in the Now Playing window.

rscullin:
This adds the **Encrypted** state to the Status column for P25 Phase 1 Voice Calls, and fixes issue #740.

I am fairly sure that this is the right way/place to fix this, and it seems to work correctly on a test recording I have.

Fixes #740 .


DSheirer:
@rscullin looks good!

PR #976, opened 2020-10-23
#815 - Change usable bandwidth percentage based on Airspy samplerate

### LINKED ISSUE #815: Airspy Mini Spectral Issues With 3.0 MHz Sample Rate
User reported issue decoding P25 control channel when using airspy mini at 3.0 MHz sample rate.  Screen shot of spectral display reflected very wide filtering drop off across the right-most 25% of the spectral display.  The poor decodes were happening when the decode channel frequency landed in this region.

Investigate if this is an issue with the airspy mini, or if sdrtrunk is not configuring something correctly fo  r the 3.0 MHz sample rate setting on the mini.

rscullin:
Percentages based on @DSheirer's percentages given in the bug report.

Tested with the Airspy Mini, and the percentages seem to be good. The usable bandwidth with Airspy Mini at 3MHz is asymmetric, but probably close enough to be fine as it is.

Fixes #815 .

### DIFF
--- a/src/main/java/io/github/dsheirer/source/tuner/TunerController.java
+++ b/src/main/java/io/github/dsheirer/source/tuner/TunerController.java
@@ -276,6 +276,14 @@ public abstract class TunerController implements Tunable, ISourceEventProcessor,
     }
 
     /**
+     * Sets the usable bandwidth percentage -- this can change based on samplerate for some tuners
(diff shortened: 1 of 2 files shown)


DSheirer:
@rscullin Nice patch.  Thank you!

Issue #888, opened 2020-07-22
Error retrieving country information from radio reference


fofeResonance:
Hi,
There was a previous issue for this that was solved with just trying several times, but that's not working for me. I'm having trouble getting the frequency information from the RR source. I'm getting this error each time with differing levels of completion.

```
12:39:15.136 ERROR i.g.d.g.p.r.RadioReferenceEditor - Error retrieving country information from radio reference - Error  [74MB/154MB 48%]
12:39:26.254 ERROR i.g.d.g.p.r.RadioReferenceEditor - Error retrieving country information from radio reference - Error  [101MB/154MB 66%]
12:39:40.714 ERROR i.g.d.g.p.r.RadioReferenceEditor - Error retrieving country information from radio reference - Error  [118MB/154MB 77%]
12:39:52.449 ERROR i.g.d.g.p.r.RadioReferenceEditor - Error retrieving country information from radio reference - Error  [95MB/154MB 62%]
12:39:55.680 ERROR i.g.d.g.p.r.RadioReferenceEditor - Error retrieving country information from radio reference - Error  [65MB/154MB 42%]
12:40:10.563 ERROR i.g.d.g.p.r.RadioReferenceEditor - Error retrieving country information from radio reference - Error  [120MB/154MB 78%]
12:40:13.754 ERROR i.g.d.g.p.r.RadioReferenceEditor - Error retrieving country information from radio reference - Error  [81MB/154MB 52%]
12:46:23.063 ERROR i.g.d.g.p.r.RadioReferenceEditor - Error retrieving country information from radio reference - Error  [126MB/154MB 82%]
12:46:25.798 ERROR i.g.d.g.p.r.RadioReferenceEditor - Error retrieving country information from radio reference - Error  [79MB/154MB 51%]
12:46:56.478 ERROR i.g.d.g.p.r.RadioReferenceEditor - Error retrieving country information from radio reference - Error  [103MB/154MB 67%]
12:47:04.315 ERROR i.g.d.g.p.r.RadioReferenceEditor - Error retrieving country information from radio reference - Error  [94MB/154MB 61%]
12:47:10.346 ERROR i.g.d.g.p.r.RadioReferenceEditor - Error retrieving country information from radio reference - Error  [71MB/154MB 46%]
12:47:13.293 ERROR i.g.d.g.p.r.RadioReferenceEditor - Error retrieving country information from radio reference - Error  [96MB/154MB 62%]
[...]

PR #992, opened 2020-10-29
RR User Interface Improvements, including Premium notices

rscullin:
This PR has some improvements around Radio Reference, Premium Accounts, and the Playlist Import system.

# Overview
- The Login pane now lets you know if you're logging in with a valid RR account, but that is _not_ a Premium account. This should help with issues like #888, where users don't realize that a Premium account is required for import.
no_premium login slash test

- The Radio Reference Playlist Importer locks out the UI elements if the account doesn't have a valid, premium account, or isn't logged in. Additionally, an error message is displayed if the account isn't Premium.
no_premium rr import v2

- The UI is enabled with a valid, Premium account.
premium_acct import

# Notes
- I added a new `SimpleBooleanProperty` for the Premium Status, as I think (?) the user can be logged in with a non-Premium account, but should still have access to Broadcastify upload.
- The PR considers an account expired if the expiration date is more than a day in the past, as I am not sure how RadioReference's timezones work. I wanted to err on the side of caution rather than locking a user out of a valid, but shortly expiring account.
- I added a JUnit test to test parsing various possible Premium Expiration values returned by the API, to ensure that my Date comparison math worked. I believe that a text string can also be returned from RR's API as the expiration date (for Admins or Broadcastify Streamers), so any API Expiration value that throws a `ParseException` is treated as a Valid Premium Subscription. The test can be run by  executing `gradle test`, or by adding the tests to IntelliJ as a Run Configuration, and probably other IDEs as well.
- If RR ever returns a "has Premium" flag in the `getUserInfo` response in a future API update, a decent part of this code can be removed, and I will be glad to remove it.
- `initRadioReferenceDecoder` is called sooner when loading systems from RR, which fixes the issue where the first time loading of a System after startup showed "Unknown" for the type of every System.

### DIFF
--- a/src/main/java/io/github/dsheirer/gui/playlist/radioreference/LoginDialog.java
+++ b/src/main/java/io/github/dsheirer/gui/playlist/radioreference/LoginDialog.java
@@ -360,6 +360,12 @@ public class LoginDialog extends Dialog<AuthorizationInformation>
                             }
                             catch(RadioReferenceException rre)
                             {
+
+                                // Set an error state
+                                getTestPassIcon().setVisible(false);
+                                getTestFailIcon().setVisible(true);
+                                getTestExpiredIcon().setVisible(false);
+
                                 if(rre.getCause() instanceof ConnectException)
                                 {
                                     Alert alert = new Alert(Alert.AlertType.ERROR, "Please check network or radio reference availability", ButtonType.OK);
@@ -391,10 +397,7 @@ public class LoginDialog extends Dialog<AuthorizationInformation>
                                 }
                             }
 
-                            // Set an error state
-                            getTestPassIcon().setVisible(false);
-                            getTestFailIcon().setVisible(true);
-                            getTestExpiredIcon().setVisible(false);
+
                             getTestConnectionButton().setDisable(false);
                         }
                     });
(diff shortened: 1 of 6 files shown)


src/main/java/io/github/dsheirer/gui/playlist/radioreference/LoginDialog.java
```diff
@@ -357,8 +391,10 @@ else if(rre.hasFault())
                                 }
                             }
 
-                            getTestPassIcon().setVisible(success);
-                            getTestFailIcon().setVisible(!success);
+                            // Set an error state
```
DSheirer, line 394: The 3 getTestXXXIcon() lines probably need moved up into the catch block.  They're overwriting what gets set within the try block.

The getTestConnectionButton() can remain after the catch block.

DSheirer (changes requested):
Great PR!  Thank you.

I corrected the one issue locally during testing and that works.  If you can make that change, I'll get this merged in.  Thanks again, Denny.

src/main/java/io/github/dsheirer/gui/playlist/radioreference/LoginDialog.java
rscullin: Oops, yeah, that needed to be moved. Fixed!

PR #1001, opened 2020-11-29
Show geo locations from DMR on map

duxet:


profporridge:
+1 for this PR. When i run this branch sdrtrunk successfully shows DMR aliases on the map

Mulderklg:
It's better to remove old trackpoints after а while. After a few hours In a high loaded network maps looks very hard to read with a lot of long tracks. The onest way to clear the map - is to restart application.

duxet:
> It's better to remove old trackpoints after а while. After a few hours in a high loaded network, map looks very hard to read with a lot of long tracks. The onest way to clear the map - is to restart an application.

That's true. I'll try to add some TTL for markers.

PR #1040, opened 2021-02-11
Sdrplay support

neilharvey94044:
Caution: These changes are dependent upon version 070 of Sammy1Am's SDRplayJava repo - implemented in a pull request I have just submitted to him.   The build.gradle reflects this dependency, so will not compile until 070 is implemented.
This set of changes adds everything Sammy1Am has in his pull request, but resolves many or perhaps all of the outstanding challenges with using SDRplay devices. 

The following features have been added:

1. Sample Size and IF Bandwidth are aligned, same as done in SDRuno from SDRplay.

2. Defaults to using IF Zero for all Sample Sizes.

3. Decimation has been implemented with allowable values based upon Sample Size.

4. Automatic Gain Control (AGC) has been implemented.

5. Defaults to DC Offset calibration enabled.

6. For RSP1A the Bias-T, FM Broadcast Notch, and Digital Audio Broadcast Notch options have been implemented.

7.  For RSP1A, the number of LNA RF Gain Reduction states matches the RSP1A capabilities.

8.  Additional SDRplay device capabilities can be more easily added in future updates.


Sammy1Am:
I've merged [the PR in the SDRplayJava library](https://github.com/Sammy1Am/SDRplayJava/pull/1), but need to figure out how to get it published to Maven Central instead of JCenter as the latter only have a few months to live apparently.  It looks like sdrtrunk is currently using jitpack.io, is this likely to stick around in the future?  It seems like an easier option for publishing artifacts for now.

Sammy1Am:
Actually it looks like jitpack.io is already working if @neilharvey94044 would make the following change to build.gradle (usernames are case-sensitive it seems):
```diff
// SDRplay support
-   implementation 'io.github.sammy1am:SDRplayJava:0.7.0'
+   implementation 'io.github.Sammy1Am:SDRplayJava:0.7.0'
```

I can still work on getting things into Maven Central eventually, but this should unblock the PR.

neilharvey94044:
I made the change you requested below.

neilharvey94044:
I've been running into challenges with the JNR library memory mapping to the SDRPlay native API.  Some of the SDRPlay API calls that were reporting no errors were in fact actually not receiving the intended data because of this.  I am continuing to work through the challenges, haven't given up yet, but in the meantime this pull request has code that looks like it's working, but actually isn't doing what is intended in some cases.  I'll submit another pull request when I'm more confident that the bridge to native code is working 100%.

neilharvey94044:
Caution: Requires SDRplayJava version 0.8.0 to compile.  Version 0.8.0 is in a pull request.

JNR struct alignment has been resolved and the following have been added.

* RSPdx features
  *  RF Notch
  *  DAB Notch
  *  Bias T
  *  Antenna Select
* IF Gain slider added - enabled when AGC is off
* RF and IF Gain sliders have been made more intuitive - slide either to right to increase total gain
* Total gain calculated by the SDRPlay API is displayed as adjustments are made
* Total gain display changes color to red when overload condition exists

Note: best results occur using AGC rather than manually adjusting IF Gain.

Sammy1Am:
FYI, SDRplayJava 0.8.0 has been tagged in GitHub, so this branch should build correctly for anyone now.

neilharvey94044:
Anyone out there reviewing pull requests?  I have additional changes to reduce GC, but they need to build on this PR.

alphafox02:
I’ve been trying anting to build this, any thoughts on it building against Ubuntu aarch64 20.04 for the pi? I’m assuming it’d treat it as arm64.

Sammy1Am:
I don't think anything here is architecture-specific.  Aside from the [jnr-ffi library](https://github.com/jnr/jnr-ffi/) (which is still pretty generic I think), everything's Java, right?  Or am I missing something obvious?

neilharvey94044:
I've built it on Ubuntu running on the Pi.  It runs, but poorly when using the SDRPlay device.  I believe it's the actual SDRPlay api doesn't seem to behave well on the Pi.  Not sure the reason.  The other purely USB devices that I have seem to work fine with this build on the Pi, however.

alphafox02:
Interesting, I’ll give it another try. I was able to build but it wouldn’t run and gave some error about a lib possibly being for x86_64 or something like that. I’ll take a closer look and paste the error. I literally just used the gradle build script except I told it where Java was on the aarch64 Ubuntu image.

alphafox02:
Thank you. That’s what I did wrong.

DSheirer:
Closing.  Resolved with #1453

PR #1083, opened 2021-09-18
Implementing TLS/SSL streaming and configuration.

whelgeson:
I don't have any streaming endpoints to actually test this on so I don't know yet if it actually works.


ImagoTrigger:
where did you come from neighbor ;p i have an icecast server i can open up for you and play with certs on it if you'd like.  my source->server connections stays on my LAN now but i'd still be interested in this if I have to use the ghetto radio relay again for some reason

PR #1085, opened 2021-09-26
Updated OpenJDK requirement

medo64:
If one tries to build project using OpenJDK 11, they will receive

    Could not target platform: 'Java SE 13' using toolchain 'JDK 11 (11)'.

Using OpenJDK 13 solves the issue.


DSheirer:
Thank you for the submission.  Closing as this is now OBE.

PR #1092, opened 2021-10-17
Properly format and safely escape to CSV in DecodeEventLogger

### LINKED ISSUE #1038: Call events log is missing a quote character, outputting malformed csv
When I enable logging of a DMR system's call events the resulting file is missing a quote character, which makes parsing the log difficult. Sample attached. The issue seems to be in the 'Timeslot' column. 
[20210204_191244.813_0_Hz_Control_call_events.log](https://github.com/DSheirer/sdrtrunk/files/5937912/20210204_191244.813_0_Hz_Control_call_events.log)

This is on 0.5.0 alpha 06

f3ndot:
Fixes #1038.

Hand-crafting a serialized format will always be a losing proposition. Brings in Apache's Commons CSV to provide a bulletproof CSV-formatted string to be written into the log. This has an additional benefit of making aliases and details safe to use quotes and commas (as well as other "reserved" CSV characters). 

To preserve the existing behaviour of SDR Trunk's CSV logging, every value/cell is wrapped in quotes, even if it isn't explicitly needed.

Adds tests to ensure the strings are as we expect.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -51,6 +51,7 @@ dependencies {
     testImplementation 'junit:junit:4.12'
     testImplementation 'org.assertj:assertj-core:3.8.0'
     testImplementation("org.junit.jupiter:junit-jupiter-api:5.7.0")
+    testImplementation 'org.mockito:mockito-core:3.+'
     testRuntimeOnly 'org.junit.jupiter:junit-jupiter-engine'
 
     //Jitpack imports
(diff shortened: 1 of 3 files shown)


DSheirer (approved):
Changes look good.

Could you please resolve merge conflicts with the (master-branch) build.gradle file and I'll merge this?

Thanks!

f3ndot:
@DSheirer welcome back, hope you're feeling rested and well.

Master merged in, builds and tests pass

DSheirer (approved):

PR #1100, opened 2021-11-14
#36 NBFM squelch control. Updates NBFM decoder to incorporate squelch…

### LINKED ISSUE #36: Create FM squelch control
Create an FM squelch control and add it to the NBFM decoder

DSheirer:
#36 NBFM squelch control. 
* Updates NBFM decoder to incorporate squelch control.  
* Squelch is a configuration item and also available to adjust at runtime with new 'Channel' tab in Now Playing window.  
* All decoders were updated to produce continuous channel power (dB) measurements that are displayed in the Channel tab's new Power Meter.
* Talkgroup identifier can be specified in decoder config.  Value is assigned to all audio segments produced to enable recording, streaming, etc.
* Playlist editor updated to include new option to specify NBFM talkgroup as an alias identifier.

Resolves #36


ah1102:
Wouldn't it work with AM modulation?

arctickitty:
im still getting severe squelch tail noise after this update sometimes the power moving as much as 2 db can knock the squelch off

PR #1112, opened 2021-11-23
Updates Latitude from LRRP packets to work correctly in southern hemisphere. 

grharrison583:
Updates Latitude from LRRP packets to work correctly in southern hemisphere.


DSheirer:
I think this issue is resolved with MR #1114, which fixed the hemisphere parsing for both latitude and longitude.  Please advise if you think there's still an issue with the way the code is parsing lat/lon values?

Issue #1133, opened 2022-01-10
PR #1129 breaks NBFM


ge0metrix:
After building the latest version of the code, NBFM no longer functions with the following:
`Exception in thread "AWT-EventQueue-0" java.lang.IllegalStateException: FM Demodulator with channel bandwidth [12500.0] requires a channel sample rate of [25000.0] - sample rate of [12631.578947368422] is not supported`

Not sure where it's getting the original sample rate to attempt to compare with, as I have the tuners rate set to 2.560MHz, and cannot find a location for setting the channel sample rate. 

#1129

DSheirer:
Resolved with PR #1134

PR #1134, opened 2022-01-10
#1133 Resolves decimation rate too small issue for NBFM decoder.

DSheirer:


ge0metrix:
Awesome, thanks for the quick turn around!

Brad-git-man:
I am wondering if someone could explain the version update procedure for SDRT?

hans109h:
If you want the latest commits you need to build from the master source code, otherwise you just need to wait until Denny builds and posts a release.  Further discussion on either ways should take place over at the Google Group.  The question has probably already been answered and folks can help more there.

Brad-git-man:
> If you want the latest commits you need to build from the master source code

What I want to know is, when there are public releases, do you have to manually update by backing up & moving your your play lists & TGs or is there some kind of auto upgrade procedure, like most programs have?

hans109h:
It's always a good idea to manually backup, but you don't have to, you can just run the new version.  Denny will usually provide info on if they is a special risk involved, such as not being able to downgrade after a new release.

PR #1148, opened 2022-03-04
Processing Enhancements and SIMD Vector Support

DSheirer:
Adds support for SIMD vector instructions which is an incubator feature in JDK 17.  Refactors most of the sample stream processing code to be more efficient.  Implements calibrations to determine optimal scalar/vector implementation based on the CPU and creates various factories to select the optimal implementation based on the calibration data.  Significant changes in the way that tuner native samples are processed, which reduces the overall JVM memory consumption.  Implements automatic DC component removal for all tuners.


Brad-git-man:
Can anyone say how to upgrade SDRT and are there any new beta versions worth upgrading to?

PR #1177, opened 2022-03-18
Fix message filtering in message tab

apankowski:
This PR fixes message filtering to the message tab.

In `master`, the `DecoderFactory.getMessageFilters` is not referenced at all. It seems that it was (probably mistakenly) deleted in commit 7eec82de "#6 DMR Decoder (#947)" which removed building of filter list in `ChannelProcessingManager`.

What I wasn't sure is if building of the filter list should happen in `MessageActivityPanel` (as in this PR) or during creation of the `ProcessingChain` (and only referenced from `MessageActivityPanel`).

Apart from that, I included fixes of small code issues spotted around the code base. I will leave comments on these describing reason for the change.


src/main/java/io/github/dsheirer/bits/BinaryMessage.java
```diff
@@ -35,29 +35,29 @@
      * Logical (ie constructed) size of this bitset, despite the actual size of
      * the super bitset that this class is based on
      */
-    private int mSize = 0;
+    private int mSize;
```
apankowski, line 38: field is set in constructor either way

src/main/java/io/github/dsheirer/bits/BinaryMessage.java
```diff
@@ -35,29 +35,29 @@
...
     private int mPointer = 0;
 
-    /**
-     * Bitset that buffers bits added one at a time, up to the size of the this
-     * bitset. 
-     *
-     * Note: the super class bitset behind this class may have a size larger 
-     * that the size parameter specified.
-     * @param size
-     */
```
apankowski, line 53: Moved dangling JavaDoc to constructor

src/main/java/io/github/dsheirer/filter/FilterEditorPanel.java
```diff
@@ -75,7 +75,7 @@ private void addFilterSet(FilterSet<T> filterSet, DefaultMutableTreeNode parent)
         List<IFilter<T>> filters = filterSet.getFilters();
 
         /* sort the filters in alphabetical order by name */
-        filters.sort((Comparator<IFilter<?>>) (first, second) -> first.getName().compareTo(second.getName()));
+        filters.sort(Comparator.comparing(IFilter::getName));
```
apankowski, line 78: build comparator from method reference

src/main/java/io/github/dsheirer/filter/FilterEditorPanel.java
```diff
@@ -98,7 +98,7 @@ private void addFilter(Filter<T> filter, DefaultMutableTreeNode parent)
     {
         List<FilterElement<?>> elements = filter.getFilterElements();
 
-        elements.sort(Comparator.comparing(FilterElement::toString));
+        elements.sort(Comparator.comparing(FilterElement::getName));
```
apankowski, line 101: order filter elements by their display name (as above) instead of synthetic `toString`

src/main/java/io/github/dsheirer/sample/buffer/IByteBufferProvider.java
```diff
@@ -25,17 +25,17 @@
 public interface IByteBufferProvider
 {
     /**
-     * Adds the listener to receive complex buffer samples
+     * Adds the listener to receive byte buffer samples
```
apankowski, line 28: match JavaDoc with purpose of the class

src/test/java/io/github/dsheirer/service/radioreference/RadioReferenceTest.java
```diff
@@ -8,7 +8,6 @@
 import java.util.Date;
 
 import static io.github.dsheirer.service.radioreference.RadioReference.CheckExpDate;
-import static org.junit.Assert.*;
```
apankowski, line 11: unused JUnit 4 import

src/main/java/io/github/dsheirer/record/wave/WaveWriter.java
```diff
@@ -105,7 +105,7 @@ public WaveWriter(AudioFormat format, Path file, long maxSize) throws IOExceptio
      */
     public WaveWriter(AudioFormat format, Path file) throws IOException
     {
-        this(format, file, Integer.MAX_VALUE * 2);
+        this(format, file, 0);
```
apankowski, line 108: because of Java type conversion and promotion rules, multiplication of `Integer.MAX_VALUE` (of type `int`) and `2` (of type `int`) will overflow (to type `int`) even though the third argument is of type `long`.

what we likely want here is to default to `MAX_WAVE_SIZE`, so we can pass 0 or just `MAX_WAVE_SIZE` if you prefer

apankowski (commented):
Added comments describing small fixes

src/main/java/io/github/dsheirer/record/wave/WaveWriter.java
ImagoTrigger: thanks maybe this will fix wav recordings (a feature i use) maybe some calls getting cut off after 8 minutes or so on long (large) files

PR #1182, opened 2022-03-20
Update build.gradle

ImagoTrigger:
Adds back the latest JavaFX gradle plugin and requests version 17 of the jfx libs in the build process

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -21,6 +21,7 @@ plugins {
     id 'java'
     id 'idea'
     id 'org.beryx.runtime' version '1.12.7'
+    id 'org.openjfx.javafxplugin' version '0.0.12'
 }
 
 repositories {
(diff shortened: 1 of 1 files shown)


ImagoTrigger:
this should help building the project when the machine has a JDK installed that does not include JavaFX

thpryrchn:
This pull request should be merged, as I wasn't able to compile on my Mac without it.

pilot51:
I'm doing regression testing for an NBFM bug introduced in v0.5.0-beta1 and this solved my build issues following 0eec2e7. Installing JDKs with JavaFX didn't help for some reason. IntelliJ saw the packages but gradle didn't.

DSheirer:
Since the build system already accounts for the JavaFX libraries and the required JDK specifies using the version with integrated JavaFX libraries, I'm going to close this PR.

PR #1213, opened 2022-04-23
Normalize audio before recording

doug-hoffman:
Normalize peak audio level in each recording to -0.44dBFS, limited to a maximum boost of ~19.5dB, to help quiet people be less quiet.

### DIFF
--- a/src/main/java/io/github/dsheirer/audio/AudioUtils.java
+++ b/src/main/java/io/github/dsheirer/audio/AudioUtils.java
@@ -24,7 +24,10 @@ import org.slf4j.LoggerFactory;
 
 import java.io.ByteArrayOutputStream;
 import java.io.IOException;
+import java.lang.Float;
+import java.lang.Math;
 import java.nio.ByteBuffer;
+import java.util.ArrayList;
 import java.util.List;
 
 public class AudioUtils
@@ -78,4 +81,43 @@ public class AudioUtils
 
         return stream.toByteArray();
     }
+
+    /**
+     * Normalize audio samples
+     */
+    public static List<float[]> normalize(List<float[]> audioBuffers)
+    {
+        float targetMax = 0.95f; // -0.44dBFS
+        float max = 0.1f; // Initialize at 0.1 (-20dBFS) to limit maximum gain
+
+        for(float[] audioBuffer: audioBuffers)
+        {
+            for(float value: audioBuffer)
+            {
+                float newMax = Math.max(max, Math.abs(value));
+                if(!Float.isNaN(newMax))
+                {
+                    max = newMax;
+                }
+            }
+        }
+
+        if(targetMax == max)
+        {
+            return audioBuffers;
+        }
+
+        float gain = targetMax / max;
+
+        for(float[] audioBuffer: audioBuffers)
+        {
+            for(int i=0;i<audioBuffer.length;i++)
+            {
+                audioBuffer[i] *= gain;
+            }
+        }
+
+        return audioBuffers;
+    }
+
 }
--- a/src/main/java/io/github/dsheirer/audio/convert/MP3AudioConverter.java
+++ b/src/main/java/io/github/dsheirer/audio/convert/MP3AudioConverter.java
@@ -18,6 +18,7 @@
  */
 package io.github.dsheirer.audio.convert;
 
+import io.github.dsheirer.audio.AudioUtils;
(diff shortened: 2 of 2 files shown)


boteman:
Would this be the place to request a drop/insert audio capability? I use VSThost in my audio processing chain with frequency equalizer and speech AGC plug-ins that do very well. This would relieve SDRtrunk of having to be everything to all users; just hand it off to software best suited to the task. TNX.

doug-hoffman:
That _could_ be another path forward, but peak amplitude normalization and an entire audio processing interface are two different levels of complexity, both in terms of implementation and runtime impact. One of the biggest problems with the latter is that Java doesn't appear to have an actively maintained VST interface.

This change wouldn't prevent you from doing any processing post-SDRtrunk – if anything, it'd make it easiest since you'd be seeing more consistent levels. Since it doesn't make any attempt to compress dynamic range or treat frequencies differently, its effect on the audio shouldn't be obvious.

boteman:
I was thinking of providing a choice in SDRtrunk to go the normal path or another sound device selectable by the user. Set the parameters, open the pipe, send audio data that would've originally been sent directly to the Icecast server; give it a detour on the way. You're already handling that audio, so just send it elsewhere.

doug-hoffman:
Using virtual audio devices to ship recordings to another application for processing would be problematic, but for other reasons.

The way audio is handled for streaming and recording, each decoded audio segment is processed individually after it ends, not as a continuous stream. Virtual audio devices operate as a continuous stream in real time. This presents two problems: 1) it would take a long time to process; 2) we wouldn't be able to easily identify the start/end of each segment on the way back.

The first problem can't be solved. We can only send the audio in real time – if it's a 30 second long segment, it'd take 30 seconds to process. What I'm doing in this PR is only limited by the speed of the CPU and will complete almost instantly.

Since the audio device would always be sending audio back, even when we aren't sending anything, how do we identify where the segment starts and ends on the way back? To continue associating metadata with each segment, they need to remain separate. Basing it on the timing of the audio we're sending would only work if there is no latency in the other application, which is unlikely. We could treat silence as a delimiter, but what if the segment has some silence in it? We could insert some pattern that would be unlikely to occur naturally, but the processing would likely alter it.

What you're proposing would be doable for the audio that already gets played back locally in real time, but I get the impression you already send that audio through another application for processing, so you know that can be done. At the point in the icecast code where we're actually dealing with audio as a stream, the audio is already in MP3 format, so we wouldn't be able to take the raw PCM from the virtual audio device and output it directly.

There is a request (issue 1071) to stream in real time, rather than waiting for a transmission to complete before processing it, which, if implemented, may make your request possible. I would suggest opening an issue with your request, if you haven't already, and referencing "# 1071" (without the space), so that they are linked and it is taken into consideration if streaming is reworked.

src/main/java/io/github/dsheirer/audio/AudioUtils.java
```diff
@@ -78,4 +81,44 @@ public static byte[] convert(List<float[]> audioBuffers)
...
+        }
+
+        if(targetMax == max)
+        {
+            return audioBuffers;
+        }
+
+        for(float[] audioBuffer: audioBuffers)
+        {
+            float[] newBuffer = new float[audioBuffer.length];
```
DSheirer, line 113: Doug, could you reuse the original float arrays and modify the samples in-place to avoid having to create new float arrays?

src/main/java/io/github/dsheirer/audio/AudioUtils.java
```diff
@@ -78,4 +81,44 @@ public static byte[] convert(List<float[]> audioBuffers)
...
+            for(int i=0;i<audioBuffer.length;i++)
+            {
+                newBuffer[i] = audioBuffer[i] * targetMax / max;
```
DSheirer, line 116: You might consider pre-calculating a gain value outside of the loops to minimize the workload:

```
float gain = targetMax / max;

for(float[] ...)
{
   for(i = ...)
   {
      audiobuffer[i] *= gain;
   }
{
```

src/main/java/io/github/dsheirer/audio/AudioUtils.java
doug-hoffman: Sure, having limited experience with java, it was unclear whether audioBuffer was a reference or a copy. I see now that it is a reference, so modifying it directly does seems like the better way to do this.

PR #1214, opened 2022-04-23
Check in with Broadcastify Calls to stay online while otherwise idle

doug-hoffman:
Runs a Broadcastify Calls test every 15 minutes to update the 'last seen' and avoid errant warnings about the node being offline.

### DIFF
--- a/src/main/java/io/github/dsheirer/audio/broadcast/broadcastify/BroadcastifyCallBroadcaster.java
+++ b/src/main/java/io/github/dsheirer/audio/broadcast/broadcastify/BroadcastifyCallBroadcaster.java
@@ -68,6 +68,8 @@ public class BroadcastifyCallBroadcaster extends AbstractAudioBroadcaster<Broadc
     private static final String MULTIPART_TYPE = "multipart";
     private static final String DEFAULT_SUBTYPE = "form-data";
     private static final String MULTIPART_FORM_DATA = MULTIPART_TYPE + "/" + DEFAULT_SUBTYPE;
+    private static final long BROADCASTIFY_TEST_INTERVAL = 15l; // minutes
+    private ScheduledFuture<?> mBroadcastifyTestFuture;
     private Queue<AudioRecording> mAudioRecordingQueue = new LinkedTransferQueue<>();
     private ScheduledFuture<?> mAudioRecordingProcessorFuture;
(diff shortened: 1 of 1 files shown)


GTR8000:
I'm not sure that I like this setting forced upon us. Frankly, I prefer to get the notification if there hasn't been any activity in X amount of hours on the node. Having the software ping the server every 15 minutes just to avoid a notification takes away my ability to keep an eye on my nodes.

Real world example of why this is not a good idea: I once had an issue with a dongle that prevented proper decoding. I got the Calls notification for the node being offline, checked SDRTrunk, and realized that it was no longer decoding that particular site.

This "feature" needs to be made optional so that those of us who have no interest in artificially keeping the connection alive can opt out of it.

davidmckenzie:
I agree with @GTR8000 on that one - would prefer it to be optional. I've had the same issue dongles screwing up and being notified through the offline notification. Prior to calls going offline from this the only notification I'd get was broadcastify issue reports.

doug-hoffman:
The node is technically up in that case. It's the SDR that's "down". The problem I see more frequently (almost daily) is a lack of activity on a system causing alerts. I either constantly get alerts that the node is down or have to set the timeout so high that it would be down for half a day before I found out (and I still got false alerts at 12 hours).

Perhaps the reoccurring test could be contingent upon the health of the system somehow? I'm not sure if there's an easy way to check that off the top of my head.

GTR8000:
The point is that if decoding stops due to any sort of issue with the dongle or antenna or any other factor, we get a notification from BCFY that the node is down, because no data has been sent in X amount of hours. Your "fix" for nagging emails on a slow node is too heavy handed, and kills our ability to have the offline notifications function as intended.

Make your enhancement optional, defaulted to the current behavior. This is not a trivial change and for some of us it's a negative change, so the default behavior should be for this option to be opt-in, not forced upon us.

doug-hoffman:
It seems the real problem is that Broadcastify is trying to identify multiple failure modes with a single test, in a way that the optimal parameters for each failure mode must be different, but yet can't be. Perhaps we could get them to have two different alerts for "last test" and "last upload", so that we may quickly identify a failed node, even if it's slow, and still offer the ability to identify issues receiving/decoding.

In #1218, I've added a configuration toggle for the scheduled connection test, defaulting to off, as well as a configurable interval that defaults to 15 minutes.

PR #1215, opened 2022-04-23
Generate audio metadata with correct to/from

doug-hoffman:
Modifies getToIdentifier() and getFromIdentifier() to return the most appropriate value, rather than the first, and then use them in AudioMetadataUtils for consistency.

### DIFF
--- a/src/main/java/io/github/dsheirer/record/wave/AudioMetadataUtils.java
+++ b/src/main/java/io/github/dsheirer/record/wave/AudioMetadataUtils.java
@@ -89,37 +89,41 @@ public class AudioMetadataUtils
 
         if(identifierCollection != null)
         {
-            for(Identifier to: identifierCollection.getIdentifiers(Role.TO))
+            StringBuilder sb;
+
+            Identifier to = identifierCollection.getToIdentifier();
+            if(to != null)
             {
-                StringBuilder sb = new StringBuilder();
+                sb = new StringBuilder();
                 sb.append(to.toString());
 
-                List<Alias> aliases = aliasList.getAliases(to);
+                List<Alias> toAliases = aliasList.getAliases(to);
 
-                if(!aliases.isEmpty())
+                if(!toAliases.isEmpty())
                 {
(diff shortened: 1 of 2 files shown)


src/main/java/io/github/dsheirer/record/wave/AudioMetadataUtils.java
```diff
@@ -89,38 +89,35 @@ public static Map<AudioMetadata, String> getMetadataMap(IdentifierCollection ide
...
-            for(Identifier from: identifierCollection.getIdentifiers(Role.FROM))
-            {
-                StringBuilder sb = new StringBuilder();
-                sb.append(from.toString());
+            audioMetadata.put(AudioMetadata.TRACK_TITLE, sb.toString());
 
-                List<Alias> aliases = aliasList.getAliases(from);
+            Identifier from = identifierCollection.getFromIdentifier();
+            sb = new StringBuilder();
+            sb.append(from.toString());
```
DSheirer, line 109: There's potential for 'from' identifier to be null here.  Please add null check/safety.

src/main/java/io/github/dsheirer/record/wave/AudioMetadataUtils.java
```diff
@@ -89,38 +89,35 @@ public static Map<AudioMetadata, String> getMetadataMap(IdentifierCollection ide
...
-            for(Identifier to: identifierCollection.getIdentifiers(Role.TO))
-            {
-                StringBuilder sb = new StringBuilder();
-                sb.append(to.toString());
+            StringBuilder sb;
 
-                List<Alias> aliases = aliasList.getAliases(to);
+            Identifier to = identifierCollection.getToIdentifier();
+            sb = new StringBuilder();
+            sb.append(to.toString());
```
DSheirer, line 96: There's potential for 'to' identifier to be null here.  Please add null check/safety.

Previously, the for/each loop protected against a null identifier because it would only loop on the non-null returned values.

PR #1218, opened 2022-04-26
Make Broadcastify Calls scheduled test configurable

doug-hoffman:
Adds a configuration toggle for the scheduled connection test, defaulting to off, as well as a configurable interval that defaults to 15 minutes.

### DIFF
--- a/src/main/java/io/github/dsheirer/gui/playlist/streaming/BroadcastifyCallEditor.java
+++ b/src/main/java/io/github/dsheirer/gui/playlist/streaming/BroadcastifyCallEditor.java
@@ -34,6 +34,7 @@ import javafx.scene.control.TextField;
 import javafx.scene.layout.GridPane;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+import org.controlsfx.control.ToggleSwitch;
 
 /**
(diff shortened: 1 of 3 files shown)


GTR8000:
Thank you.

src/main/java/io/github/dsheirer/gui/playlist/streaming/BroadcastifyCallEditor.java
```diff
@@ -177,6 +188,22 @@ protected GridPane getEditorPane()
             GridPane.setConstraints(getMaxAgeTextField(), 1, row);
             mEditorPane.getChildren().add(getMaxAgeTextField());
 
+            Label testEnabledLabel = new Label("Scheduled Connection Test");
```
DSheirer, line 191: Doug, could you rename the labels for these two additions to something that would be more intuitive to a general user, so we can avoid questions like 'what does this do'?

Since the feature seems more like a 'keep-alive' than a test, maybe something like:

'Send Periodic Keep-Alive' and 'Ping Interval'

doug-hoffman: Labels and logging updated to use new terms.

Issue #1079, opened 2021-08-20
Broadcastify Calls intermittently sends incorrect "to" identifier on DMR system


doug-hoffman:
```BroadcastifyCallBroadcaster.getTo()``` only considers the first identifier of ```Role.TO```, which isn't always the talk or patch group. On a DMR (MOTOTRBO Cap+) system, the first identifier is sometimes the radio identifier being responded to, even though the actual destination is the talk group.

Logic needs updated to be more in line with ```IcecastBroadcastMetadataUpdater.getSong()```, which looks for a patch group first, then talk group, and only then gives up and uses the first identifier. Alternatively, only using the last identifier seems to be preferred in the case of DMR, but that may not be consistently true.

doug-hoffman:
Fixed in #1227

PR #1227, opened 2022-05-02
Prefer talkgroup ID over radio ID for BCFY Calls

doug-hoffman:
Updates Calls to take advantage of new getToIdentifier() logic to prefer patch group > talk group > other identifier. 

Corrects issue observed with DMR calls having both identifiers, with a radio ID first, which was causing us to send the wrong identifier.


doug-hoffman:
Forgot to note, this resolves #1079.

PR #1236, opened 2022-05-07
Don't show tuner on spectral display while disabled

### LINKED ISSUE #1231: Disable Spectrum Not Persisted.
Disable state is not persisted across application restart.

doug-hoffman:
Prevent showTuner() from starting spectral display after user has disabled it. Persists disabled state through restarts.

Resolves #1231.


DSheirer:
@doug-hoffman Thank you.

Issue #1237, opened 2022-05-07
Actual sample rate is inconsistent with config


doug-hoffman:
Setting the sample rate above 16kHz results in a sample rate of 16kHz:

BCFY Calls:
```
$ file 1651891145-20071.mp3 
1651891145-20071.mp3: Audio file with ID3 version 2.4.0, contains:MPEG ADTS, layer III, v2,  16 kbps, 16 kHz, Monaural
```

Icecast:
```
$ mpv <url>
 (+) Audio --aid=1 (mp3 1ch 16000Hz)
File tags:
 icy-title: Scanning...
AO: [coreaudio] 16000Hz mono 1ch floatp
A: 00:00:01 / 00:00:07 (19%) Cache: 6.0s/95KB
```

doug-hoffman:
I've now seen the sample rate change between restarts without changing the configuration. It has been 22.05kHz occasionally, including right now after setting to 44.1kHz and restarting.

doug-hoffman:
Adding debug:

```java
    public MP3AudioConverter(InputAudioFormat inputAudioFormat, MP3Setting setting, boolean normalizeAudio)
    {
        mInputAudioFormat = inputAudioFormat;
        mEncoder = LameFactory.getLameEncoder(inputAudioFormat, setting);
        mNormalizeAudio = normalizeAudio;

        mLog.debug("Audio format:" + mInputAudioFormat);
        mLog.debug("Effective sample rate:" + mEncoder.getEffectiveSampleRate());
```

Confirms LAME is overriding the requested sample rate:

```
15:32:07.138 DEBUG i.g.d.a.c.MP3AudioConverter - Audio format:16-Bit 22050 Hz (default)  [41MB/80MB 52%]
15:32:07.138 DEBUG i.g.d.a.c.MP3AudioConverter - Effective sample rate:16000  [41MB/80MB 52%]
```

doug-hoffman:
I'm not seeing a clear way to fix this while using the ```LameEncoder``` class. We need to be able to call ```mEncoder.lame.getFlags().setOutSampleRate()```, which isn't accessible, to prevent it from resampling. Warning seems like the best immediate path forward.

PR #1255, opened 2022-05-20
Disable resampling if requested sample rate not supported

doug-hoffman:
If the MP3 encoder is unable to support the configured sample rate, disable resampling and warn, rather than resampling twice to the incorrect sample rate.

Warning should help users identify #1237 and self-resolve configuration, since it's not possible to disable resampling in ```LameEncoder```.

The defaults should be adjusted to 16kbps/16kHz or 32kbps/22.05kHz to have a valid combination. Let me know if you'd like me to include that in this PR or submit another, and your preference, if so. Pros and cons, both ways.


DSheirer:
Thanks again @doug-hoffman

Issue #1287, opened 2022-07-17
DMR Event Tab UI won't update in real time


Nokoa:
The event tab for DMR channels will not update in real time while clicked on a channel. Deselecting and reselecting the channel in the currently playing tab will update the events. 

I believe this issue is related to the event tab filtering. It appeared that in the DecodeEventModel.java file for DMR events fall for
`if (!mEventFilterSet.passes(event))`

DSheirer:
Issue resolved with PR #1384 by disabling event filtering until the feature is fully implemented.

PR #1099, opened 2021-11-11
Add decode event type to other DecoderStates to support filtering.

VolFFProCoder:
This is in addition to [Add filter button to messages #1098](https://github.com/DSheirer/sdrtrunk/pull/1098).

This adds the event type to allow for filtering of the remaining system protocols.

PR #1295, opened 2022-07-24
Fixes to Decode Event filtering

apankowski:
This is my attempt at fixing current implementation of decode event filtering which has the following problems:

1. Wrong layout with excessive space between button bar and decode event list view:
   image
2. Current decode event filter set not covering all `DecodeEventType`s - `CALL` was missing
3. `DecodeEventBuilder` not honoring `DecodeEventType` set on it
4. Decode event list ("Events" view) is not properly reversing list of events and filtering it using set filters when changing channels in "Now Playing" view - notice the ordering below before the fix. However, new events are added at the top.
   image

5. Most of the `DecodeEvent`s in the codebase do not have `DecodeEventType` set at all (so it's `null`). As a consequence they do not pass `EventFilter.passes` and by extension - they are not added to the "Events" list and do not receive updates (e.g. call duration). This is fixed by adding a new "Everything else" node in the filter tree which acts as a catch-all for all such events. An alternative would be to go over the whole codebase and make sure each created `DecodeEvent` has type specified.

Any remarks welcome :slightly_smiling_face: 

*Edit*: I see that #1099 is a follow-up to the #1098 that introduced event filtering. I think this PR could be an interim solution before #1099 gets merged (and fix issues that are not addressed by #1099)

*Edit 2*: I started adding some JUnit tests to verify the new behavior, following boy-scout approach. I plan to add tests for most of the changes in this PR.

*Edit 3*: I added tests following TDD and then adjusted implementation to adhere to the contract specified by the tests.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -44,11 +44,11 @@ dependencies {
 
     // JUnit Tests
     testImplementation 'junit:junit:4.12'
+    testImplementation "org.junit.jupiter:junit-jupiter-api:5.7.0"
+    testRuntimeOnly 'org.junit.jupiter:junit-jupiter-engine:5.7.0'
+    testRuntimeOnly 'org.junit.vintage:junit-vintage-engine:5.7.0'
     testImplementation 'org.assertj:assertj-core:3.8.0'
-    testImplementation("org.junit.jupiter:junit-jupiter-api:5.7.0")
     testImplementation 'org.mockito:mockito-core:3.+'
-
-    testRuntimeOnly 'org.junit.jupiter:junit-jupiter-engine'
 
     //Jitpack imports
     implementation 'com.github.dnault:libresample4j:master'
--- a/src/main/java/io/github/dsheirer/identifier/IdentifierCollection.java
+++ b/src/main/java/io/github/dsheirer/identifier/IdentifierCollection.java
@@ -20,6 +20,8 @@
 package io.github.dsheirer.identifier;
 
 import io.github.dsheirer.identifier.configuration.AliasListConfigurationIdentifier;
+import org.apache.commons.lang3.builder.EqualsBuilder;
+import org.apache.commons.lang3.builder.HashCodeBuilder;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -79,6 +81,7 @@ public class IdentifierCollection
                 mAliasListConfigurationIdentifier = (AliasListConfigurationIdentifier)identifier;
             }
         }
+        mTimeslot = timeslot;
     }
 
     public int getTimeslot()
@@ -98,14 +101,6 @@ public class IdentifierCollection
     public AliasListConfigurationIdentifier getAliasListConfiguration()
     {
         return mAliasListConfigurationIdentifier;
-    }
-
-    /**
-     * Indicates if this collection has an alias list specified.
-     */
-    public boolean hasAliasListConfiguration()
-    {
-        return mAliasListConfigurationIdentifier != null;
     }
 
     /**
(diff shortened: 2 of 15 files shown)


Nokoa:
Does this also fix #1287 ?

apankowski:
@Nokoa I think it does. Description of #1287 sounds exactly like what I've observed for all decoder types I've tested this with (it's point (5) from this PR's description). This didn't include DMR specifically, but the affected code-paths apply to all decoder types. Feel free to check if it fixes it with your setup 🙂

Nokoa:
At a first glance right now, it appears that DMR Events are appearing in real time. Thank you!

src/main/java/io/github/dsheirer/module/decode/event/DecodeEvent.java
```diff
@@ -40,7 +40,7 @@ public class DecodeEvent implements IDecodeEvent
     private Protocol mProtocol;
     private Integer mTimeslot;
 
-    public DecodeEvent(long start)
+    protected DecodeEvent(long start)
```
apankowski, line 43: this constructor is only used by builders, hence it can be `protected`

src/main/java/io/github/dsheirer/identifier/IdentifierCollection.java
```diff
@@ -332,4 +327,26 @@ public String toString()
         }
         return sb.toString();
     }
+
+    @Override
+    public boolean equals(Object o) {
```
apankowski, line 332: `equals` and `hashCode` are for tests

build.gradle
```diff
@@ -44,12 +44,12 @@ dependencies {
 
     // JUnit Tests
     testImplementation 'junit:junit:4.12'
+    testImplementation "org.junit.jupiter:junit-jupiter-api:5.7.0"
+    testRuntimeOnly 'org.junit.jupiter:junit-jupiter-engine:5.7.0'
+    testRuntimeOnly 'org.junit.vintage:junit-vintage-engine:5.7.0'
```
apankowski, line 49: add vintage JUnit engine capable of handling JUnit 4 tests

src/test/java/io/github/dsheirer/CommonFixtures.java
```diff
@@ -0,0 +1,23 @@
...
+
+import io.github.dsheirer.channel.IChannelDescriptor;
+import io.github.dsheirer.identifier.IdentifierCollection;
+import io.github.dsheirer.identifier.Role;
+import io.github.dsheirer.module.decode.dmr.channel.DMRLogicalChannel;
+import io.github.dsheirer.module.decode.dmr.identifier.DMRTalkgroup;
+
+import java.util.List;
+
+public interface CommonFixtures {
```
apankowski, line 11: this will likely evolve into per-entity fixtures. for now it's enough to have a common support interface that can be mixed in to be able to construct basic building blocks with some default values

apankowski (commented):
self review :slightly_smiling_face:

GTR8000:
I tested this out a few minutes ago and noticed that the Events message filter now has the **Everything Else** selection, which all P25 events seem to now fall under. In other words, there's currently no filtering available, it's all or nothing. I do see that you mention this in your first post here, so it's a known issue.

Is there an estimated timeline for the events to get properly mapped back to their individual filter selections?

apankowski:
@GTR8000 part 2 of solving the problems is in #1099. The author had problems with rebasing (see thread in that PR) but with his permission I could help him and (if needed) take it over and finish. In that case I think a realistic timeline would be 4 weeks from now since it's holiday season. If the author doesn't let me finish it (lets me help him with rebasing) I cannot provide any timeline as it's then not really on me.

DSheirer:
Closing.  Contents of this PR were merged under PR #1601

PR #1300, opened 2022-08-03
Fixes bitstream recording and adds some tests for changed behavior

apankowski:
Bitstream recording doesn't seem to work after some fairly recent changes as:

1. `BinaryToByteBufferAssembler` isn't flipping the buffer before handing it out to the listener,
2. `Dispatcher` used `equals` on poison pill which in case of bitstream recording is an empty `ByteBuffer`. Due to [peculiar semantics](https://docs.oracle.com/javase/7/docs/api/java/nio/ByteBuffer.html#equals(java.lang.Object)) of `java.nio.ByteBuffer#equals` the empty ByteBuffer (poison pill) is equal to a fully filled `ByteBuffer`, which immediately stopped the `Dispatcher` and in consequence - the recording.

This PR attempts to fix these issues and adds some basic tests for `BinaryToByteBufferAssembler` and `Dispatcher`

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -47,6 +47,7 @@ dependencies {
     testImplementation 'org.assertj:assertj-core:3.8.0'
     testImplementation("org.junit.jupiter:junit-jupiter-api:5.7.0")
     testImplementation 'org.mockito:mockito-core:3.+'
+    testImplementation 'org.awaitility:awaitility:4.2.0'
 
     testRuntimeOnly 'org.junit.jupiter:junit-jupiter-engine'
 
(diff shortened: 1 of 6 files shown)


DSheirer:
@apankowski thank you!  I apologize that I didn't accept your PR as requested ... I'm having trouble figuring out how to get IntelliJ to checkout remote PR's so that I can test them ... I probably just need to invest more time to figure it out.  I recreated your solution as a separate PR and merged into master.  I also updated the DibitAssembler class because it was also impacted by the ByteBuffer.flip() issue.

apankowski:
@DSheirer no problem! :)

To view PRs in IntelliJ I typically either:

- Add remote (the forking repo) & pull from it. Afterwards branches from the remote appear in Remote Branches. In particular there should be the PR branch.
- Use the [official GitHub plugin](https://plugins.jetbrains.com/plugin/13115-github) which (after configuring the GitHub token) should allow you to view and fetch pull requests via _Git_ → _GitHub_ → _View Pull Requests_.

First option is a bit more hassle but might be easier if you're used to CLI.

PR #1372, opened 2022-12-19
Disable resampling if requested sample rate not supported

doug-hoffman:
If the MP3 encoder is unable to support the configured sample rate, disable resampling and warn, rather than resampling twice to the incorrect sample rate.

Warning should help users identify #1237 and self-resolve configuration, since it's not possible to disable resampling in ```LameEncoder```.

The defaults should be adjusted to 16kbps/16kHz or 32kbps/22.05kHz to have a valid combination.

@DSheirer: Second attempt, now with the correct commit (sorry!). Fixes issue in #1255 with resampling code using the argument ```inputAudioFormat``` instead of ```mInputAudioFormat```. (I had that fixed locally but apparently forgot to push it.)

### DIFF
--- a/src/main/java/io/github/dsheirer/audio/convert/MP3AudioConverter.java
+++ b/src/main/java/io/github/dsheirer/audio/convert/MP3AudioConverter.java
@@ -57,13 +57,21 @@ public class MP3AudioConverter implements IAudioConverter
     public MP3AudioConverter(InputAudioFormat inputAudioFormat, MP3Setting setting, boolean normalizeAudio)
     {
         mInputAudioFormat = inputAudioFormat;
+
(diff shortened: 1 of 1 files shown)


DSheirer:
@doug-hoffman  since we control which InputAudioFormat and MP3Setting that the user can select, I adjusted the User Preferences dialog to disallow selecting any combinations that the LAME encoder doesn't support.  I also added a startup check to fix the user settings if they currently have a bad combo.  I incorporated your changes from this PR submission.  Thanks for digging into this and figuring out what the issue was.

Closing.  Changes incorporated into PR: https://github.com/DSheirer/sdrtrunk/pull/1376

PR #1395, opened 2023-01-02
Allow customization of icecast metadata format

doug-hoffman:
Adds the ability to customize the icecast metadata format at runtime in each stream's configuration. Default values are provided that are similar to the current behavior, with the addition of time and separation of tones from other ```FROM``` identifiers.

Also includes several related improvements in other areas to support this change, including:

- New methods in ```AudioFrames``` to get the current position as a time, get a segment of audio, get the current frames as a byte array.
- Performance improvements to icecast broadcasting that reduce overhead of metadata generation, since it's now more computationally expensive.
- New methods to get all ```TO``` or ```FROM``` identifiers contained in an identifier collection in the preferred order.
- Addition of alias list, group, site, system, and time (actual time when idle and audio time when playing back audio) as values that are available for use in metadata format.

The global logging level was reduced from ```ALL``` to ```DEBUG``` due to the verbosity of PebbleTemplates at ```TRACE```.


doug-hoffman:
The new ```getToIdentifiers()``` logic brought back #1079. I've worked around by only including the radio IDs if there are no patch/talk groups. This seemed like the desired behavior for all protocols, not just DMR, so I didn't limit to just DMR.

PR #1396, opened 2023-01-02
Allow duplicate detection by radio ID to record and stream correctly

doug-hoffman:
Adds ```TO``` identifiers (most importantly, talkgroup IDs) from duplicates to the audio that survives duplicate detection, so that it gets recorded/streamed correctly.

This PR does not address the assumption in most places about an audio segment having a single identifier. #1395 addresses that assumption for icecast metadata.


ikm19:
I have this problem and it shows in the interesting way. I would see 2 sites stream same talk group, then record one of them, but murphy's law is that it records from site farther away that brakes the recording from time to time. is it possible to make it record on both then leave just longer recording? that way which ever is broken AKA shorter is discarded?

PR #1401, opened 2023-01-06
Improvements to icecast failure handling

### LINKED ISSUE #1398: Icecast 2.3 feed gets stuck offline
During normal operation, if an icecast feed goes offline and encounters an error while attempting to reconnect (mount point in use), it'll believe it's actually connected and give up retrying.

```
2023-01-05 00:04:30.371 INFO  i.g.d.a.b.AudioStreamingBroadcaster - [East Central Emergency Network Fire and EMS] status: Disconnected  [11GB/24GB 47%]
2023-01-05 00:04:43.690 INFO  i.g.d.a.b.AudioStreamingBroadcaster - [East Central Emergency Network Fire and EMS] status: Disconnected  [8GB/24GB 37%]
2023-01-05 00:04:57.546 INFO  i.g.d.a.b.AudioStreamingBroadcaster - [East Central Emergency Network Fire and EMS] status: Disconnected  [6GB/24GB 28%]
2023-01-05 00:05:11.546 INFO  i.g.d.a.b.AudioStreamingBroadcaster - [East Central Emergency Network Fire and EMS] status: Disconnected  [6GB/24GB 25%]
2023-01-05 00:05:25.563 INFO  i.g.d.a.b.AudioStreamingBroadcaster - [East Central Emergency Network Fire and EMS] status: Disconnected  [5GB/24GB 24%]
2023-01-05 00:05:29.680 INFO  i.g.d.a.b.AudioStreamingBroadcaster - [East Central Emergency Network Fire and EMS] status: Connected  [9GB/24GB 39%]
```

image

The change below in #1381 seems to indicate the new behavior is intentional, but it is not very desirable for temporary problems. Perhaps it should convey the error without preventing connection attempts.

src/main/java/io/github/dsheirer/audio/broadcast/icecast/IcecastTCPAudioBroadcaster.java:

```java
            //Only set broadcast state to disconnected if it's not already in an error state, to prevent a restart.  We
            //want to preserve the error state that got us here, so the user can see it.
            if(!getBroadcastState().isErrorState())
            {
                setBroadcastState(BroadcastState.DISCONNECTED);
            }
```

Making a change to the feed then saving allows it to reconnect successfully.

doug-hoffman:
- Add mechanism to persist failure reason while continuing to reconnect
- Add 'warning' attribute for failure conditions that may be temporary, but still need to be reported
- Add logging for failures
- Prevent possibly temporary server-side conditions from blocking reconnect attempts (fixes #1398)
- Detect icecast 2.4 auth failures
- Prevent generic errors from overriding existing error state

image

image

### DIFF
--- a/src/main/java/io/github/dsheirer/audio/broadcast/AbstractAudioBroadcaster.java
+++ b/src/main/java/io/github/dsheirer/audio/broadcast/AbstractAudioBroadcaster.java
@@ -34,6 +34,7 @@ public abstract class AbstractAudioBroadcaster<T extends BroadcastConfiguration>
     private Listener<BroadcastEvent> mBroadcastEventListener;
     private T mBroadcastConfiguration;
     protected ObjectProperty<BroadcastState> mBroadcastState = new SimpleObjectProperty<>(BroadcastState.READY);
+    protected ObjectProperty<BroadcastState> mLastBadBroadcastState = new SimpleObjectProperty<>();
     protected int mStreamedAudioCount = 0;
     protected int mErrorAudioCount = 0;
     protected int mAgedOffAudioCount = 0;
@@ -56,6 +57,14 @@ public abstract class AbstractAudioBroadcaster<T extends BroadcastConfiguration>
     }
 
     /**
+     * Observable last bad broadcast state property
+     */
+    public ObjectProperty<BroadcastState> lastBadBroadcastStateProperty()
+    {
+        return mLastBadBroadcastState;
+    }
+
+    /**
      * Current state of the broadcastAudio connection
      */
     public BroadcastState getBroadcastState()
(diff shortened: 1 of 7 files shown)


DSheirer:
@doug-hoffman Thanks for the PR.  

Comment:  streaming errors are going to be temporary and once they're addressed, they shouldn't be a long-term issue.  The primary sdrtrunk gui's streaming status panel is already kinda busy with several columns and adding another column to that table seems to make it more crowded. 

Do you think you can achieve the same effect (capture the last error to support troubleshooting) by only including the new column in the playlist editor's streaming configuration table?  The primary status panel will show that the feed has some kind of error status which should be a cue to the user to go to the playlist editor to diagnose the issue further.

doug-hoffman:
Thanks for the feedback. I've dropped the changes to the streaming status panel, so now the last error will only be displayed in the playlist editor.

davidflypei:
It would be great to get this update. Is there anything else that needs to be done so that other broadcasters don't get locked in an error state?

doug-hoffman:
I don't think this fixes #1892. I still have issues with feeds to icecast 2.4 breaking after a network disruption, but it did fix my issues with previous versions of icecast. I've not spent much time troubleshooting the 2.4-specific issue.

PR #1407, opened 2023-01-15
#1397 Decoding Channels Go IDLE After Period of Time

### LINKED ISSUE #1397: Control Channel Decode Goes to IDLE After Period of Time
This is a continuing issue that was not resolved in version 0.5.0 final release.

When running 3x control channels, at some point all 3 channels go to an IDLE state and nothing is being decoded.  I suspect that in each case the signal is being tuned/mixed outside of the ability of the PLL to pull the signal back in.

Stopping and restarting just one of the channels causes the 2x remaining channels to spring back to life and start decoding again.  This would indicate that the channelizer is resetting the mixer offset for the 2 remaining channels and causing the channel mixer to pull the signal back to baseband.

Previous work under #1353 added thread locking to ensure that multi-threaded channel access didn't cause the tuner and provided channels to get out of sync.  This may have been an underlying issue, but doesn't seem to be the primary cause of the symptoms described in this issue.

This issue was observed with the Polyphase channelizer.

# Troubleshooting
1. Implement a channel FFT view and add it to the 'Channel' tab and include the channels current PPM and Mixer offset values to determine what the channel is being commanded to do and detect why that's different that what's needed to baseband the signal.
2. Track the mixer & PPM values over time, along with decoder state to identify when the signal is mixed out of baseband and what were the values being applied to the PPM and mixer offset at the time.

DSheirer:
Closes #1397 

Resolves issue where decoding channels go IDLE after period of time.  Isolated cause to a multi-threading issue where the polyphase channel manager receives notification that the tuner center frequency has changed and sets a flag to update each of the channels.  This happens concurrent with a previous tuner center frequency change where the flag gets set and the second thread is updating the channel output processors with the previous frequency change and then unsets the frequency change flag, completely missing the second/latest frequency change update and causing the channels to all tune-out their desired signals as the tuner center frequency changes and the output channels continue mixing to the previous frequency.

Identified and resolved issue where a 2-stream polyphase synthesizer was being updated with an incorrectly designed synthesis filter which was causing massive droop on the second half of the reconstructed channel.  Oddly enough, this didn't seem to impact the P25 decoders, but it may have impacted other decoding.

Identified and resolved issue with concurrent modification of the modules list in the processing chain associated with recent addition of the channel FFT view in the channel power tab, with a previous PR under this same issue.

### DIFF
--- a/src/main/java/io/github/dsheirer/dsp/filter/channelizer/ChannelCalculator.java
+++ b/src/main/java/io/github/dsheirer/dsp/filter/channelizer/ChannelCalculator.java
@@ -1,31 +1,34 @@
-/*******************************************************************************
- * *********************************************************************************************************************
- * sdr-trunk
- * Copyright (C) 2014-2017 Dennis Sheirer
+/*
+ * *****************************************************************************
+ * Copyright (C) 2014-2023 Dennis Sheirer
  *
- * This program is free software: you can redistribute it and/or modify it under the terms of the GNU General Public
- * License as published by  the Free Software Foundation, either version 3 of the License, or  (at your option) any
- * later version.
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
  *
- * This program is distributed in the hope that it will be useful,  but WITHOUT ANY WARRANTY; without even the implied
- * warranty of  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General Public License for more details.
+ * This program is distributed in the hope that it will be useful,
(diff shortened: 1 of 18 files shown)


Waymorehoss:
i'm having problems with this my control channel will stay constant for a good while then it will stop and be idle after about 15 seconds it will go back to constant...does that mean something is wrong is it suppose to do that ?

DSheirer:
This issue is resolved, so unless you're using an older version of sdrtrunk, it shouldn't be for this specific issue.

What you're describing sounds like your signal may be too weak and it may be fading in an out, or there may be a stronger signal nearby that's impacting it.  

You can repost your question in one of the support forums if you need assistance in resolving it.

Waymorehoss:
ok one other thing. i do think it's the signal strength that's causing it the control channel also will break in the waterfall feed do control channels do that take a break every now and then?

PR #1582, opened 2023-06-24
#1581 Rdio Scanner Streaming Configuration Editor - Host URL Path

### LINKED ISSUE #1581: RdioScanner Should Auto-Configure Host URL Path
The RdioScanner streaming editor requires the end user to type the full endpoint path after the host which causes confusion for most users.

Update the editor and the streaming code to automatically include/add the path to the hostname provided by the user.

DSheirer:
Closes #1581 
Updates Rdio Scanner configuration editor to automatically append the 'api/call-upload' path to the end of the protocol and host value provided by the user.  Sets the host default value to 'http://localhost' but allows the user to overwrite.


tadscottsmith:
Thank you for correcting this!

DSheirer:
> Thank you for correcting this!

You're welcome!

PR #1677, opened 2023-10-21
#1607 P25 P1/P2 L3Harris Patch Group Support

### LINKED ISSUE #1607: Harris P25 Patching Behavior
I have a local Harris P25 system that I monitor, they are in the process of migrating from EDACS to P25. Now the system is still in the very early stages of migrating. The EDACS system has a big habit of having all kinds of TG patches done daily. And both UniTrunker and Trunking Recorder handle them greatly. 

But moving onto P25 and SDRTrunk, since it is a Phase 2 System and UT Cant decode the audio...On the active call window in SDR Trunk and Trunking recorder will only show me the patch TG ID, which is in the 651xx range, and won't show me the Supper and Subgroups that are being patched with the 651xx talk group. 

Further, Harris P25 systems that use Symphony consoles, which this system is using, use a feature called Simulselect when a dispatcher "patches" two talk groups together, which creates a third dynamic talk group in the 65xxx range.  (Similar to how EDAC creates a 14-xxx or 15-xxx TG). The 65xxx range talk groups are members of a pool, meaning that they are not permanently assigned to any particular agency or talk group pair; they are created and torn down depending on demand.

This is a known Harris P25 obstacle. The Super and Subordinate talkgroups are not being displayed because Harris doesn't send out a Voice Patch Group Grant over the control channel as Motorola does. Motorola announces explicit patch calls. This is true both for 3600 baud Motorola trunking as well as Motorola's P25 offering. This means Motorola has the ability to issue a non-patched group call while a patch exists for that group ID, but Harris doesn't. Instead, it just sends out a plain old Group Voice Grant message for the Supergroup, leaving out any mention of it being a patch call. So as expected, so then SDR Trunk displays it as a normal group call without showing the subgroups in the display. EDACS (4800, 9600, and SitePro) uses implicit patch calls - meaning the presence of one or more patches imply the call is a patched group call. Before announcing a non-patched call on these systems the patch must be dropped. Otherwise, all subscribers will assume it is a patched call. 

When monitoring a Hariss P25 Systems the expected behavior is to show the Patched TG as well as the Supper and Subgroups that are being patched, Like this. (I can't reproduce a screenshot from SDRTrunk only TR) 
image

Instead, only the 65xxx range talk groups show up, like this. 
image
So this means that there is no way to tell where the patch call is actually coming from unless you are monitoring the system with something else to show it patches like Unitrunker. I, unfortunately, cannot capture an active screenshot in SDRTrunk as the patches are not active currently not active but when I can ill reply to this thread 

**Conclusion: Since this is a Harris P25 issue is it possible that SDRTrunk can handle Harris P25 Systems more like Motorola and/or EDACS?**

DSheirer:
Closes #1607

Add support for L3Harris dynamic patch/regroup messages and events.  
Updates patch group manager to include support for patch groups of individual radios.  
Includes parsing for newly discovered L3Harris unknown TSBK opcodes: 10, 42 and 43.


josephdambrosio:
Had some time to look at the behavior when patching.  It looks like the same issue is still happening with not following the TGs being patched and classified as the TG they're patched into.  Attached is a ZIP of short period of recordings when there was a patch active to a mutual aid county.  

System TG 2350 is patched using the Harris Dynamic Patch TG of 65132
[pathrec.zip](https://github.com/DSheirer/sdrtrunk/files/13382709/pathrec.zip)

PR #1724, opened 2023-11-11
#1635 Add a special county agency to show all county freqs without an agency

### LINKED ISSUE #1635: RadioReference import doesn't show all county systems
**Is your feature request related to a problem? Please describe.**
In RadioReference, there are county systems in the database that aren't shown within the RR import feature.  For example, "Local Schools" from this page are not shown in SDRTrunk:

https://www.radioreference.com/db/browse/ctid/2048#cid-5850

**Describe the solution you'd like**
The ability to select all systems within the RR database.

**Additional context**
RR post that details the issue:

https://forums.radioreference.com/threads/sdrtrunk-rr-conventional.462058/

EricTendian:
Closes #1635 

This solves the problem of frequencies tied to the main county page in RadioReference not showing up in the list of frequencies available to import. I'm doing this via a new subclass called `CountyAgency` which acts like an agency, but gets the frequency categories from the county instead of an agency. It will always show up first in the UI, like so:

image

For counties with a very large number of categories (and likely subcategories), it will default to showing the first category in the list, instead of trying to show all the frequencies at once (which I've found to be extremely slow):

image

On other smaller counties though, it will have the normal `(All)` selection.

### DIFF
--- a/src/main/java/io/github/dsheirer/gui/playlist/radioreference/AgencyEditor.java
+++ b/src/main/java/io/github/dsheirer/gui/playlist/radioreference/AgencyEditor.java
@@ -23,6 +23,7 @@ import io.github.dsheirer.playlist.PlaylistManager;
 import io.github.dsheirer.preference.UserPreferences;
 import io.github.dsheirer.rrapi.type.Agency;
 import io.github.dsheirer.rrapi.type.AgencyInfo;
+import io.github.dsheirer.rrapi.type.CountyInfo;
 import io.github.dsheirer.service.radioreference.RadioReference;
 import io.github.dsheirer.util.ThreadPool;
 import javafx.application.Platform;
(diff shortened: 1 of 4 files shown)


GTR8000:
I just tested this with the latest source code of beta5 and found a glitch, otherwise it seems to be working well.

- Navigate to County A, then navigate to County B, then return to County A. **X County (All)** gets duplicated each time you bounce back and forth between counties (an additional entry for each is added at the top of the list).

EricTendian:
@GTR8000 Thanks for testing it! You're right, it looks like I forgot that Java will pass the original object along, so it kept adding the CountyAgency to the original list of agencies in the API. Now I'm effectively copying the list and don't see the same issue anymore.

GTR8000:
Yup, looks good. @DSheirer tested with very latest Beta 5 source code, working well, recommend merging ASAP

DSheirer (approved):
Great submission, thank you!

PR #1730, opened 2023-11-12
Fix Rdio Scanner API call to comply with HTTP standards

EricTendian:
I was running into some weird issues with my proxy that was processing Rdio scanner API calls, it turns out that SDRTrunk doesn't send a valid HTTP request, so not all servers can process it correctly. There cannot be two content type headers, so I'm removing the `audio/mpeg` one as it's able to be derived from the `.mp3` filename in Rdio scanner itself. Additionally, there was an extra space in the multipart form filename syntax, which may also cause some issues.

I've verified that the changes work fine with Rdio scanner.


tadscottsmith:
Thanks Eric!

PR #1792, opened 2024-01-11
P25-P2 Encryption Synchronization ReSync

crsawyer:
- Enhanced resync of ESS processor
- Re-Order of AbstractVoiceTimeslot processing
- Updated P25 MBE recorders to include synchronized encryption information (mainly to verify this fix worked)

Resync:
When a channel goes from "Call->Hangtime->PTT->Call", the first call is not guaranteed to end on a Voice-2 (See TIA-102.BBAC, Figure D-5). When the second call initiates, the ESS processor does not always start on the first Voice-4. Updated to reset ESS processor on PTT, END_PTT, HANGTIME, and IDLE.

Re-Order:
On a Voice-2 slice, the computed ESS applies to the following Voice slices. Re-ordered so that the current Voice-2 is processed before the computed ESS is processed.


DSheirer:
Thanks!

Issue #1763, opened 2023-12-14
In presence of multiple SDRs choose preferred ones


rxes:
**Is your feature request related to a problem? Please describe.**
I have multiple types of SDRs, Nooelec, SDR RTL, AirSpy Discovery, SDRPlay RSPa. When all plugged in, it always defaults to Nooelec, which is not optimal. Seems that even when SDRPlay or AirSpy Discovery, is selected as default tuner for control, it's never used to pick up talk, but Nooelec gets to do it.

**Describe the solution you'd like**
Ability to select preferred tuners that can decode the voice calls

**Describe alternatives you've considered**
None

**Additional context**
![403405912_938604477674046_8051809460258304623_n](https://github.com/DSheirer/sdrtrunk/assets/153829270/190d55a6-b200-4fd9-a80f-8cfb97f4f12f)
![403402143_733007198235157_8879632593452111391_n](https://github.com/DSheirer/sdrtrunk/assets/153829270/a9aee3b1-0b9e-40b4-ada1-a9401262b97e)
![403402000_1038070867521198_4310795786998559753_n](https://github.com/DSheirer/sdrtrunk/assets/153829270/93b0cd84-15ef-4f5d-ade8-59a0a9d19f9d)

DSheirer:
Closing as resolved under https://github.com/DSheirer/sdrtrunk/pull/1808

PR #1808, opened 2024-01-25
Inherit P25 Trunked Preferred Tuner To Traffic Channels

Nokoa:
While monitoring a P25 Trunked system, the control channel will follow the preferred tuner conguration, however all of its traffic channels will not. They may use any Tuner.

The preferred tuner value is null for all the traffic channels, as they are not inherited from the parent.

TunerManaer received a null preferred tuner string value.

I manually set the preferred tuner from the parent in all references on the P25TrafficChannelManager class

This allows all voice channels associated with the P25 site to attempt to first use the preferred tuner.

Please test if this works

#1763

### DIFF
--- a/src/main/java/io/github/dsheirer/module/decode/p25/P25TrafficChannelManager.java
+++ b/src/main/java/io/github/dsheirer/module/decode/p25/P25TrafficChannelManager.java
@@ -310,7 +310,9 @@ public class P25TrafficChannelManager extends TrafficChannelManager implements I
                     SourceConfigTuner sourceConfig = new SourceConfigTuner();
                     sourceConfig.setFrequency(frequency);
                     if(mParentChannel.getSourceConfiguration() instanceof  SourceConfigTuner parentConfigTuner)
+                    {
                         sourceConfig.setPreferredTuner(parentConfigTuner.getPreferredTuner());
+                    }
                     trafficChannel.setSourceConfiguration(sourceConfig);
                     mAllocatedTrafficChannelMap.put(frequency, trafficChannel);
 
@@ -359,7 +361,9 @@ public class P25TrafficChannelManager extends TrafficChannelManager implements I
             SourceConfigTuner sourceConfig = new SourceConfigTuner();
             sourceConfig.setFrequency(frequency);
             if(mParentChannel.getSourceConfiguration() instanceof  SourceConfigTuner parentConfigTuner)
+            {
                 sourceConfig.setPreferredTuner(parentConfigTuner.getPreferredTuner());
+            }
             trafficChannel.setSourceConfiguration(sourceConfig);
             mAllocatedTrafficChannelMap.put(frequency, trafficChannel);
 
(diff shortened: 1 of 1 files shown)


GTR8000:
This would definitely have benefits in some scenarios, but could potentially be problematic in other scenarios.

For example a mixed 700/800 site where the control channel is 800 and some of the traffic channels are 700. Obviously no single tuner is going to cover 769-860. So what happens when the 800 control channel has Tuner 1 set as preferred tuner, and there's a voice grant on 700 and the traffic channel tries to inherit the preferred tuner of the 800 control channel? Is it going to throw an error that the tuner is out of range, or would it select another tuner that is able to decode the 769-775 range? And if the latter, is it quick enough not to miss activity?

Nokoa:
From what I understood of it. It is just the preferred tuner. If tuner 1 is set as preferred and used for 800 mhz and is locked, should it receive a call grant for 700mhz, I believe it will first try tuner 1, see that the frequencies are not in range, and that it cannot be retuned, before moving to the next tuner. Tuner 1 will not be forced to be used. I don't think that this affects that logic. Because traffic channels prior had no preference at all, and they may or may not go in this order of trying irrelevant tuners anyhow.  

I have 2 tuners plugged in Airspy and RTL-SDR. My entire P25 system can be received fully on the Airspy. The Airspy also has the proper antenna equipped for that system, yet SDRTrunk will tune some channels on the Airspy but also utilize the RTL-SDR, and tune significant amount of channels there too, even thought it doesn't at all need to. It feels quite random. Or I am missing something.

Right now though, I do see several issues with tuning that can be improved also. Tuners' frequencies get changed while running even though the entire bandwidth of the system is well within range.

Or in Cap+ with 2 frequencies, SDRTrunk may retune the center frequency as it's moving between channel 1 and channel 2, even though they can be received at the same time well within the bandwidth of the SDR.

I believe some more logic can be implemented as a whole for tuning, one which can keep track of all the trunked voice frequencies, and adapt the SDRs center frequencies as needed to cover all of them. However for people who don't have the hardware available to cover all the frequencies all the time, this may cause an issue. 

I believe when a preferred tuner is selected this was the expected behavior, that the entire site attempt to use the preferred tuner first.

GTR8000:
I agree that the tuning logic is sometimes...well, illogical.

I have a RSP2pro running @ 10 MHz (8.5 MHz usable) decoding a few P25 sites in the 769-775 block. Obviously more than enough bandwidth to handle the entire band. Still, I had to create "dummy" channels on 768 and 776 to keep the heart of the target spectrum centered in order to avoid dropoff at the edges, because SDRTrunk wanted to keep re-tuning the center frequency. Actually quite frustrating, and being able to manually lock a center frequency would eliminate the need to create those silly "dummy" channels to prevent re-tuning when it's wholly unnecessary.

Anyway as I said, what you've done will definitely be beneficial in some cases where a single tuner can handle all of the traffic frequencies associated with the site's control channel, and frankly any attempt to fine tune SDRTrunk's handling of spectrum and center channels and tuning is most welcome. Thanks!

src/main/java/io/github/dsheirer/module/decode/p25/P25TrafficChannelManager.java
```diff
@@ -309,6 +309,8 @@ private void processPhase1ChannelGrant(APCO25Channel apco25Channel, ServiceOptio
                     broadcast(event);
                     SourceConfigTuner sourceConfig = new SourceConfigTuner();
                     sourceConfig.setFrequency(frequency);
+                    if(mParentChannel.getSourceConfiguration() instanceof  SourceConfigTuner parentConfigTuner)
+                        sourceConfig.setPreferredTuner(parentConfigTuner.getPreferredTuner());
```
DSheirer, line 313: Could you please wrap each of the sub-clause additions in braces.  I know Java allows single-line sub-clause without braces, but it can lead to logic/readability errors.

PR #1901, opened 2024-04-29
#1876 Unifies Reentrant Lock Between Tuners and Frequency Controller.

### LINKED ISSUE #1876: Unify Tuner & Frequency Controller Locks to Avoid Thread Deadlock
**sdrtrunk Version**
master

**Describe the bug**
RTL2832 based tuners use a ReentrantLock to control thread access to the hardware to deconflict automated processes and the user's ability to make minor changes.  All tuners also employ a reusable FrequencyController to manage the currently tuned center frequency and sample rate and notifying consumers of changes, and this controller also employs a ReentrantLock.  This can cause a thread deadlock when a tuner is disabled and the user enables the tuner again at the same time that a multi-frequency channel is rotating frequencies.  Attached diagnostic log shows the two threads (21 and 67) in a dead-locked state where the channel is trying to get a new frequency channel source while the tuner is still in the process of initializing during restart.  This is a rare use-case, but the existence of two locks within the same resource sets the stage for other deadlock scenarios.

1. Update the Tunable interface with a method to access the tunable's resource lock.
2. Update the base tuner controller to expose the ReentrantLock.  The lock is currently in the RTL2832TunerController, but should be pushed up to the TunerController parent class to comply with the Tunable interface.
3. Update the FrequencyController to remove the ReentrantLock and instead use the lock from the Tunable member variable.

**Application Log**
[20240313_115607_sdrtrunk_processing_diagnostic_report.log](https://github.com/DSheirer/sdrtrunk/files/14646721/20240313_115607_sdrtrunk_processing_diagnostic_report.log)

DSheirer:
Closes #1876

### DIFF
--- a/src/main/java/io/github/dsheirer/source/tuner/frequency/FrequencyController.java
+++ b/src/main/java/io/github/dsheirer/source/tuner/frequency/FrequencyController.java
@@ -22,11 +22,12 @@ import io.github.dsheirer.source.ISourceEventProcessor;
 import io.github.dsheirer.source.InvalidFrequencyException;
 import io.github.dsheirer.source.SourceEvent;
 import io.github.dsheirer.source.SourceException;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
 import java.util.ArrayList;
 import java.util.List;
 import java.util.concurrent.locks.ReentrantLock;
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
 
 public class FrequencyController
 {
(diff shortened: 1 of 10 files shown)


src/main/java/io/github/dsheirer/source/tuner/frequency/FrequencyController.java
```diff
@@ -57,14 +53,14 @@ public FrequencyController(Tunable tunable)
...
 
-    /**
-     * Lock for controlling access to frequency control plane, event processor subscriptions and source event broadcasts.
-     */
-    public ReentrantLock getFrequencyControllerLock()
-    {
-        return mLock;
-    }
-
+//    /**
```
DSheirer, line 56: Remove dead code

Issue #1619, opened 2023-07-31
View dialog windows open minimized


hans109h:
**sdrtrunk Version**
master-branch (6.0-alpha7)

**Describe the bug**
Playlist editor, DMR recording viewer, icon manager, and user preferences dialog windows open as completely minimized (vertical black bar) after one of them has been opened and then closed.

**To Reproduce**
1.  Start SDRTrunk
2. Open one of the dialog windows via the view menu.
3. Close the dialog window.
4. Open any other dialog window via the view menu.

**Expected behavior**
Window should open in default size and location with window contents visible.

**Screenshots**
![image](https://github.com/DSheirer/sdrtrunk/assets/5613188/9168d69c-0461-4a8a-8c3d-c2e1d8225be1)

**Application Log**
Nothing useful.

**Desktop (optional - complete the following information):**
 - OS: Ubuntu 23.04
 - CPU Cores: 8
 - RAM: Lots
 -
**Additional context**
It's been doing this for a while, I haven't had a chance to see if I can find the regression.

DSheirer:
@hans109h and @mwsltn  I'm unable to reproduce this error. 

However, I did find a code error in the JavaFxWindowManager where it wasn't monitoring the Java FX window size and position correctly.  It is supposed to monitor the windows so that it can record where the user last placed the window so that on restoration it will resize and position the window at the same place.

I'm going to submit that change against this issue and that will close out this issue.  If the change doesn't resolve this issue, please reopen this issue and we'll try again.

hans109h:
@DSheirer  unfortunately it didn't resolve the issue.  I can confirm that it does indeed remember the location (and size) of the window, so that when initially opened it returns there, but any opening of the dialog window after the initial result in the window completely minimized and only visible as a thin line on the screen.

DSheirer:
Reopening.  Will attempt to force the window out of the 'iconified' state before invoking show and focus.

DSheirer:
@hans109h I merged some additional changes to attempt to fix this issue.  The nightly build process just completed generating a new build with this change.  Can you please access the nightly build and verify if the issue is resolved?

hans109h:
Negative, issue still present.  Window opens normal on first open, after being closed it only opens in its iconified state.  If location of window is changed, the location is accurately remembered when opened even in iconified state.

hans109h:
On any open of the Playlist Editor, getting this in the syslog:
[...]

PR #2047, opened 2024-10-27
#2037 JavaFX Windows Can't Be Shrunken

### LINKED ISSUE #2037: SDRTrunk Window Scaling wont work on playlist editor
**sdrtrunk Version**
SDRTrunk Nightly (2024-09-22) , v0.6.1-beta-2, and 0.6.0 stable

**Describe the bug**
I use SDRTrunk on a Linux machine that I rarely have physical access to, so everything is done via remote. With that said, I noticed that regardless of physical or remote access, the SDRTrunk window is horribe at automatically scaling, or I cant resize it at all, even manually. So i run into situations where I actually have to resize different boxes and columns in the playlist editor just to even see the stuff I need to adjust. And this is even with a 1920x1080 monitor with 100% scaling on local and remote screens, but the window is still oversized. 

The main window opens and scales just fine, its the playlist editor thats the issue. The elements within the Playlist Editor window are fixed size and will not allow the window to be less than 1921x1568. (as reported by "xdotool selectwindow getwindowgeometry") And I also noticed that the maximize/window button between the close and minimize button does nothing for playlist editor either. Neither makes it bigger or smaller. 

Issue occurs at 1600 x 900 (16:9) aspect radio. By changing monitor resolution to 1280 x 1024 (5:4), the "drag to resize" line in the playlist editor will be visable and can be clicked and dragged to resize ONLY the elements in the page, not the window. This is somewhat of a temporary solution. But the Playlist Editor window will exceed display dimensions when screen resolution s 1600x900.

**To Reproduce**
1. Run SDRTrunk
2. Open Playlist Editor
3. Attempt to click and drag window edges to resize, and try clicking maximize / minimize buttons. 

**Expected behavior**
Neither of the actions above should resize window due to the bug.

**Screenshots**

**Application Log**
If applicable, include either the full application log,or a snippet of the error logging.

**Desktop (optional - complete the following information):**
 - OS: MX Linux v6.1.0-18 (AMD64) [Debian 6.1.76-1] 64bit. 
 - Dell OptiPlex 7010 - Intel Core i7-3770S @3.1ghz - 8GB DDR3 1600MHz.

**Additional context**
SDRTrunk group discussion regarding this... https://groups.google.com/g/sdrtrunk/c/5ZIksGBCeL8

DSheirer:
Closes #2037 
JavaFX Windows Can't Be Shrunk.  

Reverts changes made under #1619 that was applying a min width and height to the window based on the last time that the user resized the window.  This had the effect of allowing windows to get bigger, but never allowing windows to get smaller than the last width/height values on previous application shutdown.

### DIFF
--- a/src/main/java/io/github/dsheirer/gui/JavaFxWindowManager.java
+++ b/src/main/java/io/github/dsheirer/gui/JavaFxWindowManager.java
@@ -460,9 +460,6 @@ public class JavaFxWindowManager extends Application
         stage.show();
         stage.requestFocus();
         stage.toFront();
-        stage.setMinHeight(stage.getHeight());
-        stage.setMinWidth(stage.getWidth());
-        // System.out.println("Stage [" + stage.getTitle() + "] restored - width[" + stage.getWidth() + "] height [" + stage.getHeight() + "]");
     }
 
     @Override


hans109h:
Initial testing doesn't appear to be a regression.

PR #2098, opened 2024-12-02
Add Motorola OTA alias decoding

ilyacodes:


DSheirer:
Thank you Ilya!  Impressive!  I'll get this tested and merged.

GTR8000:
I've been running the modified file against a few ASTRO 25 systems with known aliases, and so far I'm impressed with the results of the decoding. Well done, sir!

Issue #2153, opened 2025-02-19
Motorola Talker Aliases Identifier Support


tadscottsmith:
**sdrtrunk Version**
Latest nightly.

**Describe the bug**
Motorola Talker Aliases are not being added as an identifier.

**Expected behavior**
Motorola Talker Aliases should be added as an identifier.

tadscottsmith:
Resolved by #2154

PR #2154, opened 2025-02-19
Add Motorola Talker Alias Identifier

tadscottsmith:
This should correct #2153.

Handles MotorolaTalkerAliasComplete in the same way that LCHarrisTalkerAliasComplete is currently handled.

Feel free to modify or cancel if you'd like.

### DIFF
--- a/src/main/java/io/github/dsheirer/module/decode/p25/phase1/P25P1DecoderState.java
+++ b/src/main/java/io/github/dsheirer/module/decode/p25/phase1/P25P1DecoderState.java
@@ -319,7 +319,6 @@ public class P25P1DecoderState extends DecoderState implements IChannelEventList
         }
         else if(iMessage instanceof MotorolaTalkerAliasComplete talkerAlias && talkerAlias.isValid())
         {
-            mTrafficChannelManager.getTalkerAliasManager().update(talkerAlias.getRadio(), talkerAlias.getTalkerAlias());
             processTalkerAlias(talkerAlias);
         }
         else if(iMessage instanceof LCHarrisTalkerAliasComplete talkerAlias)
--- a/src/main/java/io/github/dsheirer/module/decode/p25/phase1/P25P1DecoderState.java
+++ b/src/main/java/io/github/dsheirer/module/decode/p25/phase1/P25P1DecoderState.java
@@ -317,14 +317,31 @@ public class P25P1DecoderState extends DecoderState implements IChannelEventList
                     break;
             }
         }
-        else if(iMessage instanceof MotorolaTalkerAliasComplete tac && tac.isValid())
+        else if(iMessage instanceof MotorolaTalkerAliasComplete talkerAlias && talkerAlias.isValid())
         {
-            mTrafficChannelManager.getTalkerAliasManager().update(tac.getRadio(), tac.getAlias());
+            mTrafficChannelManager.getTalkerAliasManager().update(talkerAlias.getRadio(), talkerAlias.getTalkerAlias());
+            processTalkerAlias(talkerAlias);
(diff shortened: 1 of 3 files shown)


src/main/java/io/github/dsheirer/module/decode/p25/phase1/P25P1DecoderState.java
```diff
@@ -317,16 +317,32 @@ public void receive(IMessage iMessage)
...
+    private void processTalkerAlias(MotorolaTalkerAliasComplete talkerAlias)
+    {
+        Identifier identifier = getIdentifierCollection().getFromIdentifier();
+
+        if(identifier instanceof RadioIdentifier radioIdentifier)
+        {
+            mTrafficChannelManager.getTalkerAliasManager().update(radioIdentifier, talkerAlias.getTalkerAlias());
+        }
+
+        mTrafficChannelManager.processP1TrafficCurrentUser(getCurrentFrequency(), talkerAlias.getTalkerAlias(), talkerAlias.getTimestamp(), talkerAlias.toString());
```
DSheirer, line 343: @tadscottsmith the change to pull the radio identifier from the identifier collection might not be as accurate as using the radio identifier that gets transmitted in the alias message, moreover, the radio ID might not have been captured yet, whereas it's consistently included in the talker alias complete message.  Did you find in your testing that the talker alias manager wasn't being updated in the existing implementation?

The last line that you added where you update the traffic channel manager makes sense, to get that alias added to the event.  Can you reduce the change set for this PR to just this line?

tadscottsmith: @DSheirer 

In the current implementation, it seems the talker alias manager is being updated correctly. Active System Radio Aliases are being displayed in the Details tab of each control channel, but the talker alias identifier is not making it's way to the the AudioSegment, and therefore isn't an identifier that can be consumed by the broadcasters. Calling `mTrafficChannelManager.processP1TrafficCurrentUser(...)` when the full alias has been received adds the missing identifier. 

I reverted the code to the bare minimum as you suggested. This discards all of the cosmetic changes to align the Motorola implementation with the LC/Harris implementation. (E.g. Motorola method name remains `getAlias()` while the LC/Harris implementation is `getTalkerAlias()`).

DSheirer:
I think there might need to be a lookup that happens when the current call is underway and the link control tells us who the FROM radio ID is.  I'll see if I can add some more changes to ensure this lookup happens and then the alias should make its way into the audio segment consistently afterward.

DSheirer:
I merged in your changes.  If you find that there are still issues with getting the talker alias into the audio segment, please let me know and I'll see if there's something that's preventing that from happening.

I dug into the code and the talker alias lookup for subsequent calls should be happening in P25P1DecoderState.processLCChannelUser() and as it gets added to the decoder state's identifier collection that same update should be echo'ed to the audio module for inclusion in the current audio segment.

tadscottsmith:
This seems to be working as expected now. Alias identifiers are visible to the broadcasters. I closed my incident. Thank you again for your help! 

```
2025-03-10 16:23:44.253 DEBUG i.g.d.a.b.o.OpenMHzBroadcaster - Alias: SPPD-43952  [4GB/8GB 58%]
2025-03-10 16:23:49.255 DEBUG i.g.d.a.b.o.OpenMHzBroadcaster - Alias: SPFD-P-M24A  [2GB/8GB 29%]
2025-03-10 16:25:28.259 DEBUG i.g.d.a.b.o.OpenMHzBroadcaster - Alias: SPPD-566816  [1GB/8GB 18%]
```

PR #2159, opened 2025-02-23
 DMR VoiceMessage: Fix incorrect frame assignment in DMR voice frame reconstruction

modificaz:
## Summary
Fixes an incorrect assignment to `frame_3[4]`, which should have been `frame_2[4]` based on the intended logic of the code.

## Reasoning
The extracted byte should be assigned to `frame_2[4]` to correctly reconstruct the middle 72-bit voice frame. The assignment to `frame_3[4]` leaves `frame_2[4]` untouched from the initial `0` value , thus corrupting 8 bits of crucial FEC parity and voice parameters.

## Impact
This fix should noticeably improve ECC performance and voice quality for DMR voice payloads.

## Issue Reference
No relevant issues found.


DSheirer:
Awesome find!  Thanks very much for the PR!

Issue #2198, opened 2025-03-20
Clicking on a DMR traffic channel generates exception in log


profporridge:
**sdrtrunk Version**
master-branch

**Describe the bug**
An error will be logged in the application log when clicking on a DMR traffic channel: 
```
java.lang.NullPointerException: Cannot invoke "io.github.dsheirer.module.decode.dmr.DMRTrafficChannelManager.getTalkerAliasManager()" because "this.mTrafficChannelManager" is null
	at io.github.dsheirer.module.decode.dmr.DMRDecoderState.getActivitySummary(DMRDecoderState.java:1493)
	at io.github.dsheirer.channel.details.ChannelDetailPanel.receive(ChannelDetailPanel.java:115)
	at io.github.dsheirer.channel.details.ChannelDetailPanel.receive(ChannelDetailPanel.java:38)
	at io.github.dsheirer.sample.Broadcaster.broadcast(Broadcaster.java:128)
```
	

**To Reproduce**
Monitor a DMR network. When you see traffic channels appear in the channel list, click on a channel named 'T-Control'. 

**Expected behavior**
No error 

**Screenshots**

![Image](https://github.com/user-attachments/assets/77dca7c3-42fc-492c-b57b-305641eb9741)

DSheirer:
Closing.  Resolved in #2202

PR #2199, opened 2025-03-20
fix for 2198 null mTrafficChannelManager

profporridge:
Check channel type first. Resolves https://github.com/DSheirer/sdrtrunk/issues/2198


DSheirer:
Thanks for the PR.  This change would prevent the aliases from showing when you click on the control channel.

I added a fix under #2203 to resolve this NPE issue and ensure the talker alias summary can still be accessed via either a control or traffic channel.

Closing as resolved under #2202

PR #2207, opened 2025-03-21
#2207 OpenMHz Broadcast Enhancements

tadscottsmith:
Adds patched Talkgroup and Talker Alias information to the OpenMHz broadcaster.


DSheirer:
Thanks for the PR!

PR #2263, opened 2025-07-29
attempt to persist preferred spectral and control panel sizes

n2ygk:
I have a tiny 7" LCD display and I found I had to keep sliding the spectrum display. The geometry was not being remembered. I messed around with a few things and probably did this incorrectly as I was being coached by ChatGPT since I'm not a Java developer.

This change does appear to work so I don't know if there's some magic about `setPreferredSize` vs `setSize`. I also move this code outside the `if(dimension != null)` which is probably just wrong.


n2ygk:
@DSheirer please take a look. Thanks.

Issue #2216, opened 2025-04-13
Feature - Display alias list in channels tab


dave3825us:
Any chance a column can be added in the Play List editor (Channels Tab) to display the alias list associated with each system? Sort of like this?

![Image](https://github.com/user-attachments/assets/d8074a7f-c037-4ad5-a61c-f9ee841d55a9)

And have it clickable to allow sorting like when clicking on Systems, Sites, Freqs etc? 

Thanks

DSheirer:
This should be fairly easy to add.

PR #2270, opened 2025-08-09
Display alias list in channels tab

TimmySheff:
Addresses Feature Request #2216


fourohthree:
@TimmySheff can you split the commits for fixing Noise Squelch Panel and the Alias List into two PRs? I think the former commit is being overlooked as a result

TimmySheff:
I believe Dennys last update made  changes to this area so I have closed
that PR.

I am totally new to this GitHub process.  I will re look at this with
Dennys new code and resubmit if necessary.

Issue #2243, opened 2025-06-24
MBE Recordings Should Include Additional Metadata


DSheirer:
**Is your feature request related to a problem? Please describe.**

As requested here: https://github.com/lwvmobile/dsd-fme/issues/298#issuecomment-2994414382

@DSheirer would if be possible to add an MFID string to the JSON formatted .mbe files in the future in SDRTrunk? Reason needed, is that on some encryption, its the only differentiation between how to decrypt (i.e., DMRA/Moto BP vs Hytera BP).

Edit: Also, would it be feasible to add system identifiers, such as NAC, SYSID, RFSS ID, WACN and so on for P25, if available, and CC and Trunking SYSCODE on DMR if available? Or too much work?

lwvmobile:
Also, just a followup to the MFID string or value, if placed in the .mbe file, very specifically for DMR, Ideally want the MFID of either:

1. Voice Link Control Frames prior to call, particularly if it is Hytera.
2. The MFID of the PI Header (if it is in the start of the call sequence).

If a PI header is present, it can supersede the MFID of the VLC frames, since it signals an enhanced privacy, and not basic.

Unfortunately, DMR Encryption is in such a state that you have to check down for at least 4 different possibilities, as opposed to just looking at one ESS field for P25. The possibilities include:

1. No PI Header, But SVC Options Signalling Encrypted Voice.
2. No Pi Header, But SVC Options Signalling Encrypted Voice and MFID 0x68 signalling Hytera.
3. PI Header Present, MFID 0x10 signalling DMRA Enhanced Privacy Modes (and SVC Options signalling enc).
4. PI Header Present, MFID 0x68 signalling Hytera Enhanced Privacy Modes  (and SVC Options signalling enc).

And that sadly, doesn't even cover the late entry mechanisms that both DMRA and Hytera use throughout a call.

Cagier:
I can get those error rates down as that was a quick test on a secondary antenna. I hadn't thought about the embedded IV so that is a really clever solution! (I would expect no less from you! 😉)

Are you using a modified version of DSD-FME with the -c option or did you create a standalone script for this? (I can probably produce something myself to do the same but it will take me a while to work out the details.)

Where the algo is predetermined and there is more than one (known) key then I guess trying the various keys and confirming which one produces the expected VC1 value would allow successful identification of the KeyID as well.

[...]

PR #2273, opened 2025-08-14
P25 P1 MBE ESS order

crsawyer:
In the recorded MBE files ESS was populated at the beginning of LDU2, but doesn't take effect until after LDU2 is processed. This increases complexity of post processing.

- Corrected to now populate ESS after LDU2
- Also added 'version' field to MBE file and set it to '2' to signify a breaking change in the ESS order
- This partially satisfies items pointed out in both
  - #2243
  - lwvmobile/dsd-fme#298

I should have caught this in #1792, my bad.

### DIFF
--- a/src/main/java/io/github/dsheirer/audio/codec/mbe/MBECallSequence.java
+++ b/src/main/java/io/github/dsheirer/audio/codec/mbe/MBECallSequence.java
@@ -36,7 +36,7 @@ import java.util.List;
  * radio identifiers.
  */
 @JsonRootName("mbe_call")
-@JsonPropertyOrder({"protocol", "call_type", "from", "to", "encrypted", "frames"})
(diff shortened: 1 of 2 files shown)


lwvmobile:
You're doing The Lord's work with this one. I don't suppose you have any sample .mbe files with this format update by chance?

crsawyer:
[20250823_230521_123456789_1_705_37789_encrypted.mbe.txt](https://github.com/user-attachments/files/21955235/20250823_230521_123456789_1_705_37789_encrypted.mbe.txt)
I live about 100 miles from the closest Phase1 system. But here's what I can from running a previously recorded *.bits file through the P25P1MessageFramer.

lwvmobile:
Thanks for that, I appreciate it.

Issue #1892, opened 2024-04-19
Icecast stream gets stuck in a 'Connecting' state after a network interruption


redrob:
**sdrtrunk Version**
nightly 2024-02-13T09:27:56

**Describe the bug**
Icecast stream gets stuck in a 'Connecting' state after a network interruption or similar. There is no automatic timeout function that forces a full connect reset. For example, the network was interrupted last night at 19:42 but was restored minutes later. The stream stopped at 19:42 and never reconnected on its own. I had to manually disable and re-enable the stream. 

**To Reproduce**
1. Add icecast stream (server running Icecast 2.4.0-kh15)
2. Make sure it is in 'Connected' status.
3. Induce a momentary network interruption (ethernet/WiFi disconnect, etc)
4. Restore network connection 

**Expected behavior**
stream should automatically timeout after a certain period, completely reset and attempt a reconnect. 

**Application Log**

```
20240418 194224.089 [NioProcessor-74] ERROR i.g.d.a.b.i.IcecastHTTPAudioBroadcaster - Broadcast error  [462MB/724MB 63%]
java.net.SocketException: Connection reset
	at java.base/sun.nio.ch.SocketChannelImpl.throwConnectionReset(Unknown Source)
	at java.base/sun.nio.ch.SocketChannelImpl.read(Unknown Source)
	at org.apache.mina.transport.socket.nio.NioProcessor.read(NioProcessor.java:378)
	at org.apache.mina.transport.socket.nio.NioProcessor.read(NioProcessor.java:47)
	at org.apache.mina.core.polling.AbstractPollingIoProcessor.read(AbstractPollingIoProcessor.java:519)
	at org.apache.mina.core.polling.AbstractPollingIoProcessor.access$1200(AbstractPollingIoProcessor.java:68)
	at org.apache.mina.core.polling.AbstractPollingIoProcessor$Processor.process(AbstractPollingIoProcessor.java:1224)
	at org.apache.mina.core.polling.AbstractPollingIoProcessor$Processor.process(AbstractPollingIoProcessor.java:1213)
	at org.apache.mina.core.polling.AbstractPollingIoProcessor$Processor.run(AbstractPollingIoProcessor.java:683)
	at org.apache.mina.util.NamePreservingRunnable.run(NamePreservingRunnable.java:64)
	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(Unknown Source)
[...]

PR #2279, opened 2025-08-30
#1892 Icecast 2.4+: Fix retries after network disruption

ajcoll5:
This addresses #1892 where Icecast 2.4+ would get stuck after a network disruption.

I’m having unrelated issues with my RTL-SDRs not locking on to P25 frequencies with the latest code base, so I haven’t been able to extensively test this against main. However, I’ve been running these changes with a 6.1 snapshot for the past few weeks, and they’ve been reliable in making sure Icecast 2.4+ reconnects after network disruption.


ajcoll5:
> During testing, the changes will log every connection attempt and fail, every 30 seconds. Can you modify this to only log the connection attempt/failure once and suppress subsequent loggings?

Done!

PR #2344, opened 2026-01-04
Add CTCSS decoder and alias support

msullivan1993:
- Adds CTCSS/PL decoder to SDRTrunk
- Detects all 43 standard CTCSS tones (67.0-254.1 Hz)
- Uses Goertzel algorithm for efficient detection
- Integrates into existing alias system used by DCS for streaming

Full disclaimer.. most of this work was done by Claude Opus. I am not a coder by any stretch of the word. Tested extensively on my personal server with no issues with send to Rdio Scanner server. Unknown if any changes need done for Calls, OpenMHz, etc.

### DIFF
--- a/src/main/java/io/github/dsheirer/module/decode/ctcss/CTCSSDecoder.java
+++ b/src/main/java/io/github/dsheirer/module/decode/ctcss/CTCSSDecoder.java
@@ -90,6 +90,14 @@ public class CTCSSDecoder extends Decoder implements IRealBufferListener, Listen
     private CTCSSCode mPendingCode = null;
     private static final int MIN_CONSECUTIVE_DETECTIONS = 2;
 
+    // === NEW: Tone loss threshold - require consecutive misses before clearing ===
+    private int mToneLostCount = 0;
+    private static final int TONE_LOST_THRESHOLD = 2;
+    
+    // === NEW: Periodic rebroadcast for late-starting audio segments ===
+    private long mLastBroadcastTime = 0;
+    private static final long REBROADCAST_INTERVAL_MS = 1000; // Re-broadcast every 1 second
+
     static
     {
         FIRFilterSpecification specification = FIRFilterSpecification.lowPassBuilder()
@@ -220,7 +228,7 @@ public class CTCSSDecoder extends Decoder implements IRealBufferListener, Listen
         }
     }
 
-    /**
+ /**
      * Calculate power at each frequency and detect tone
      */
     private void calculatePowerAndDetect()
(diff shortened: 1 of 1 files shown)


Randesign:
> 
> I noticed this as well. I was working on a fix that adds Alias Matching (tone squelch, basically) but it kind of broke decoding for me on a few PL's when I shortened the block time up, I haven't had a chance to roll back the changes since I've been in bed the last few days sick (I've been using a test build that was before those changes)
> 
> I'll probably work on it tonight and see if I can get it to work better. Again, I'm no coder, most of the work is being done with Claude AI, but this takes some load off Dennis and fixes known issues that I and others have had.
> 
> As for the GUI.. yes, I saw that too, but if you go to another PL and back, it selects just fine. Probably a bug that's been there since Dennis added DPL originally. I'll try to make a fix for that too while I'm at it.
> 
> As for the non-matching alias.. since you have everything in one Alias list, it's grabbing an alias that matches that PL. The only fix I found for that was that I had to create separate Alias lists for each individual channel (royal PITA I know) since the Alias list appears to be not an "AND" function, more of an "OR" function (it will pass audio to that TG with any of the added aliases) it does make it difficult. (Maybe add that as a pull request too? That might fix it)
> 
> I considered adding this as a channel decode feature as well, just haven't started working on it yet. I might do that as well so there is both options.

I have never used the DCS decoder (there is nothing around here that uses it), so I'm not familiar with the signal flow and associated logic. 

It seems to me that the Alias is the best place to set the tones as opposed to Channel. There are cases here where the US Forest Service operates a half dozen repeaters on the same VHF frequency, with each repeater using a different PL tone. Ideally I'd only want to monitor/record two of those repeaters, so I guess setting up 2 different aliases would be the approach. Those two aliases would be in the same alias list. 

I see the problem with trying to keep the logic flow similar to a trunked radio system. It would probably necessitate separate alias lists with each channel.  On the other hand, one needs to assign a NBFM channel to a unique talkgroup so it seems like a collection of channels is acting like a virtual trunked radio system feeding a single alias list. I haven't looked at the code, so I don't know what the best approach would be without causing a massive re-write to the logic.

Randesign:
@msullivan1993 Are you still working on this? I'm thinking of writing my own version of the CTCSS implementation.

msullivan1993:
> @msullivan1993 Are you still working on this? I'm thinking of writing my own version of the CTCSS implementation.

Yes, I finally got around to working on it tonight. Just made a new PR https://github.com/DSheirer/sdrtrunk/pull/2380 to combine this and channel-level CTCSS into one PR. 

closing this PR

PR #2345, opened 2026-01-05
Add P25 NAC alias support for streaming

msullivan1993:
Adds P25 NAC (Network Access Code) alias support, allowing users to create aliases based on NAC values for streaming and identification. Tested with P25 conventional channels streaming to Rdio Scanner with correct talkgroup override. 

- Added NAC as an alias identifier type
- NAC values (0-4095) can be configured in the alias editor under P25 → NAC
- Streaming works correctly with "Stream this Alias as Talkgroup" setting
- Fixed RdioScannerBroadcaster to check all identifiers for StreamAsTalkgroup, not just the TO identifier

Files Changed
- NEW: alias/id/nac/Nac.java
- NEW: gui/playlist/alias/identifier/NacEditor.java
- Modified: 7 existing files (see commit)

Tested with Rdio scanner, unknown if any changes are needed for Calls, OpenMHz, etc. Again, I am no coder, Claude Opus did the bulk of the work.

### DIFF
--- a/src/main/java/io/github/dsheirer/audio/broadcast/rdioscanner/RdioScannerBroadcaster.java
+++ b/src/main/java/io/github/dsheirer/audio/broadcast/rdioscanner/RdioScannerBroadcaster.java
@@ -434,45 +434,48 @@ public class RdioScannerBroadcaster extends AbstractAudioBroadcaster<RdioScanner
      * Creates a formatted string with the TO identifiers or uses a default of zero (0)
      * 
      */
-    private String getTo(AudioRecording audioRecording)
+private String getTo(AudioRecording audioRecording)
+{
+    Identifier identifier = audioRecording.getIdentifierCollection().getToIdentifier();
+    AliasList aliasList = mAliasModel.getAliasList(audioRecording.getIdentifierCollection());
+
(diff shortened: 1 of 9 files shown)


src/main/java/io/github/dsheirer/audio/broadcast/rdioscanner/RdioScannerBroadcaster.java
```diff
@@ -434,45 +434,48 @@ private static String getTalkerAlias(AudioRecording audioRecording)
      * Creates a formatted string with the TO identifiers or uses a default of zero (0)
      * 
      */
-    private String getTo(AudioRecording audioRecording)
-    {
-        Identifier identifier = audioRecording.getIdentifierCollection().getToIdentifier();
+private String getTo(AudioRecording audioRecording)
```
DSheirer, line 437: The proposed changes should follow/align with the code formatting from the surrounding class.  The modified method is (now) fully left-justified and not tab indented like the original.  Additionally, use a blank line in between methods ... the proposed change removes the existing blank line between the getTo() method and the method that follows it.

DSheirer (changes requested):
Request minor code formatting changes.

Functional testing complete.

PR #2354, opened 2026-01-25
Add channel squelch Delay Time with optional silence removal for NBFM channels

msullivan1993:
- Adds "Squelch Delay" timer to NBFM channels
- Allows user to set a time to look for additional radio traffic once squelch (plus Alias) has closed before sending audio to stream
- Functions similarly to the Delay Hold feature on traditional scanners
- Adjustable from 0-5000ms in 100ms increments
- Optional "Remove Silence" will remove silence between transmissions in same clip, or can be left in for accurate timing
- Alias agnostic

Disclaimer: Another Claude Opus specialty. Works great in testing and should fix issues with intermittent dropouts due to local transmissions and prevent shortened clips

### DIFF
--- a/src/main/java/io/github/dsheirer/audio/AudioModule.java
+++ b/src/main/java/io/github/dsheirer/audio/AudioModule.java
@@ -262,10 +262,12 @@ public class AudioModule extends AbstractAudioModule implements ISquelchStateLis
                         // If not removing silence, add silence for the time squelch was closed
                         if(!mSquelchDelayRemoveSilence && mSquelchClosedTimestamp > 0)
                         {
-                            long silenceDuration = System.currentTimeMillis() - mSquelchClosedTimestamp;
+                            long elapsedMs = System.currentTimeMillis() - mSquelchClosedTimestamp;
+                            // Round up to nearest 100ms to match delay increments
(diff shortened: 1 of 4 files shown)


DSheirer (changes requested):
If I understand correctly, you're trying to detect when two audio transmissions happen close together with a (small) break in the middle, and then force the audio segments to be one continuous audio segment with an optional, fixed length segment of silence inserted in between to represent a variable length gap between the transmissions.

The current behavior causes those segments to be two separate audio segments.  

The NoiseSquelch component manages squelch control and resampling (to 8 kHz) of the demodulated audio and then sends the resampled audio to the AudioModule.  The audio module responds to the squelch control commands and produces AudioSegments of the audio for downstream recording, playback and streaming.

The proposed changes extend (ie leak) squelching control out of the NoiseSquelch and into the AudioModule where it becomes a shared responsibility.

Recommend reworking your changes to isolate the proposed changes to the NoiseSquelch control where you can implement the delay behavior and overwrite the noise audio representing the break with simple float zero values, before resampling, and before passing the resampled audio buffer to the AudioModule.  This approach preserves the clear separation of function that's in the current implementation.

I'm going to close this PR.  Please resubmit a new PR once you're able to restructure the code changes.

PR #2357, opened 2026-02-02
Adds non standard DCS code and "Select Tone" placeholder

msullivan1993:
- Adds non standard DCS code D212N
- Adds "Select Tone" default placeholder when adding new DCS identifier


DSheirer:
Thanks for the PR!

Issue #2301, opened 2025-11-17
AM airband reception issues


Radiotech012:
Noticing massive amounts of  distortion on AM/Airband signals.

Seems to impact stronger signals more than weaker ones, but demodulated AM audio is unintelligible.

NFM & P25 audio is great 👍 

Tuner is RTL dongle.

SdrTrunk version: Nightly 17/11/25

**sdrtrunk Version**
Indicate which sdrtrunk version you are using.  Use 'master-branch' if you are running the latest code.

**Describe the bug**

**To Reproduce**

**Expected behavior**

**Screenshots**

**Application Log**
If applicable, include either the full application log,or a snippet of the error logging.

**Desktop (optional - complete the following information):**
 - CPU Cores: [e.g. 4 cores]
 - RAM:[e.g. 16GB]

**Additional context**

Randesign:
Even weak signals were distorted for me. Looking at the spectrum display, it seems like the noise floor is elevated, too. It's like there is some gain being added. I haven't had a chance to look at the code. 

Were you able to get the squelch to work? It seems like it's disabled on the two versions I tried.

Radiotech012:
I suspect there are a few issues going on here.

The selected (AM) tuner Gain control MUST be set to Automatic. Any other manual gain control setting will result in severe distortion because it appears to completely eliminate any AM AGC control (which is critical). Even in Automatic Gain mode, I've still heard distortion on AM signals - But it's generally not as bad. This was a trap for me because I generally like to use manual gain settings for FM & digital modes, because it reduces the displayed spectrum noise 'floor' to a less stupid level and often results in optimised sensitivity and s/n on weak FM and digital signals.

The AM squelch control doesn't seem to operate correctly in Auto Track or manual mode. In fact, if you select Auto Track in the Playlist editor and then go back to the Channel in Now Playing, the Auto Track function is disabled (unticked).

A bigger problem, is that if you are monitoring a trunking network and AM (possible FMN channels too???) Airband channels, SDRTrunk will only play audio from one source at a time. If a trunk channel becomes active, it will block audio from the AM channel, and vice versa. It seems there is no ability to demodulate or route audio from both sources to the Left/Right speakers, or even mix the two sources into the mono channel.

Randesign:
[...]

PR #2384, opened 2026-03-06
Bugfix-#2383-audio-trapped-in-output-buffer-of-resampler

### LINKED ISSUE #2383: NBFM audio samples trapped in buffer after squelch closes
**sdrtrunk Version**
Master branch

**Describe the bug**
After a NBFM transmission completes and the squelch engages, audio samples are left in the output buffer of RealResampler. The number of samples remaining is between 1 and 511, because the BufferManager only broadcasts audio in 512 sample blocks. The samples remain in the output buffer until the next NBFM transmission, where new incoming audio is added to the buffer to bring it over the 512 threshold. 

This has the effect of prepending the leftover audio from the earlier call onto the beginning of the new call.

**To Reproduce**
I was able to verify this by injecting some audio "markers" into the un-resampled audio stream at the beginning and end of squelch closings and could clearly see in the .wav file the end of the previous transmission at the beginning of the next.

I am currently testing a fix and will issue a PR when completed.

Randesign:
Fixes #2383.

Resampler already implements a lastBatch parameter, which will pad the unused portion of the input buffer with zeros. Extended that function to the output buffer to be sure no audio samples are left in the buffer. 

The RealResampler is used by a number of functions in SDRTrunk.

Tested with:
MP3 resampling
NBFM/noiseSquelch decoding

AM demodulator will be left as is for now. I would like to address resampler issues and other issues in #2301 at a later date.

AFSK1200: I have no way to test this.

### DIFF
--- a/src/main/java/io/github/dsheirer/dsp/filter/resample/RealResampler.java
+++ b/src/main/java/io/github/dsheirer/dsp/filter/resample/RealResampler.java
@@ -40,6 +40,7 @@ public class RealResampler
     private Listener<float[]> mResampledListener;
     private BufferManager mBufferManager;
     private double mResampleFactor;
+    private boolean mLastBatch;
 
     /**
      * Constructs an instance.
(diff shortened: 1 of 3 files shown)


Randesign:
I had some corruption with git on my dev system, so I think the last two commits are duplicates.

PR #2387, opened 2026-03-12
Fix RTL-SDR and USB tuner detection on macOS without sudo

RCGV1:
## Summary

Fixes USB tuner detection and initialization on macOS without requiring `sudo`. This was tested on macOS 15 Sequoia (Apple Silicon / aarch64) with an RTL-SDR Blog V4 dongle, where the same device already worked with Homebrew `rtl_test` as a normal user.

## Root causes and fixes

**Device enumeration and matching when macOS reports port 0 / empty port chains**
- macOS IOKit reports `port = 0` for USB devices, so the old `port > 0` guards excluded devices from discovery and hot-plug handling.
- Fixed by removing the `port > 0` filter in `TunerManager`, `FCDTunerController`, and `USBTunerController.findDevice()`.
- When macOS reports an empty port chain for every device on a bus, matching by bus + port string alone is ambiguous. The shared USB controller path now validates the expected `TunerClass` from the device descriptor before selecting a device.
- `TunerFactory` now passes the expected tuner class into each `USBTunerController`, so the same matching logic applies consistently to RTL-SDR, Airspy, Airspy HF+, HackRF, and HydraSDR.

**`LIBUSB_ERROR_ACCESS` on `claimInterface`**
- `setAutoDetachKernelDriver(true)` enables `USBInterfaceOpenSeize` on macOS IOKit, which causes `LIBUSB_ERROR_ACCESS` when no kernel driver holds the interface.
- Fixed by skipping `setAutoDetachKernelDriver` on macOS.

**`LIBUSB_ERROR_PIPE` after redundant `setConfiguration(1)`**
- Calling `setConfiguration(1)` when the device is already in configuration 1 causes macOS to re-enumerate the device, which leaves endpoint 0 stalled and breaks later control transfers with `LIBUSB_ERROR_PIPE`.
- Fixed by reading the current configuration first and skipping the call when the device is already configured.

**Shared libusb context**
- `TunerManager` discovers devices in one libusb context, but the old startup path opened tuners in a second context. On macOS that caused the second open path to misbehave and later control transfers to fail.
- Fixed by sharing `TunerManager`'s libusb context with `USBTunerController`.

**RTL-SDR startup robustness**
- The RTL startup path now initializes the baseband before reading EEPROM so the I2C bus is ready.
- The macOS startup path skips the reset probe sequence that is unnecessary after a fresh macOS open.
- Descriptor access now guards against null/empty EEPROM data instead of throwing `NullPointerException` / `IndexOutOfBoundsException`.

## Validation

- [x] `./gradlew compileJava`
- [x] RTL-SDR Blog V4 detected and initialized on macOS 15 Sequoia (Apple Silicon) without `sudo`
- [ ] Verify on Linux
- [ ] Verify on Windows
- [ ] Verify hot-plug behavior on macOS with additional USB tuner hardware
- [ ] Verify Airspy / Airspy HF+ / HackRF / HydraSDR on macOS

### DIFF
--- a/src/main/java/io/github/dsheirer/source/tuner/TunerFactory.java
+++ b/src/main/java/io/github/dsheirer/source/tuner/TunerFactory.java
@@ -130,6 +130,7 @@ import java.util.ArrayList;
 import java.util.List;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+import org.usb4java.Context;
 
 import javax.sound.sampled.TargetDataLine;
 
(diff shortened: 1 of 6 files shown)


DSheirer:
@RCGV1 can you reopen and complete this?  I think these are valuable enhancements.

cheers,
Denny

RCGV1:
Reopened and completed the branch.

I also pushed a follow-up commit (`c9b0d69f`) that hardens the empty-port-address matching path by validating the expected `TunerClass` from the USB descriptor instead of using an RTL-only vendor check. That keeps the shared macOS discovery path consistent across the other USB tuner controllers too.

Validation on my side:
- `./gradlew compileJava`
- RTL-SDR Blog V4 on macOS 15 Sequoia (Apple Silicon) without `sudo`

I updated the PR description to reflect the final behavior and remaining cross-platform validation items.

RCGV1:
GitHub still refuses to reopen this PR because the stored head repository reference is detached ( via the API). I opened replacement PR #2401 from the same updated branch so review can continue there.

RCGV1:
GitHub still refuses to reopen this PR because the stored head repository reference is detached (the API reports headRepo as null). I opened replacement PR #2401 from the same updated branch so review can continue there.

PR #2404, opened 2026-04-11
Clear frequency band map when switching P25 control channels

### LINKED ISSUE #2298: SDRTrunk using incorrect frequencies with newest P25 decoder
**sdrtrunk Version**
Master branch, latest Nightly released on 11/16/25

**Describe the bug**
On a handful of P25 trunked channels, the software is not displaying the correct frequencies. For some reason it is either shifting them up a little, or in some cases a LOT. For example, one channel on the WV SIRN (Missouri Branch, Wayne Co) is programmed with the correct control channels.. however the software is displaying it as ABOVE the UHF-T band, which I know is not correct. And as I was capturing a screenshot, the Milton control channel was showing 20GHz.. earlier today, one of the Point Pleasant control channels was showing as 12GHz. Because of this, I assume the software is also attempting to open voice channels that are not correct, and traffic is being missed.

Now, if I remove the non-active control channel and restart the channel, everything seems to appear normal. I don't know if it has to do with having multiple CC's in the list or not but I will have to find some more examples.

**To Reproduce**
Unknown if it can be reproduced. Only 3 of around 20 control channels presenting this behavior with no logic, although all three problem channels were on the WV SIRN system.

**Expected behavior**
Control channels to show correctly and voice channels to activate correctly.

**Screenshots**
Pic 1: Channel config with the two control channels for the Missouri Branch site.
Image

Pic 2: SDRTrunk showing the control channel as 515.637MHz, along with inaccurate primary and secondary control channel info. Also notice the frequency of the Milton site.
Image

Pic 3: After removing the inactive control channel and restarting the channel, all is working as expected.
Image

Pic 4: From earlier in the day, the same problem was popping up on the Point Pleasant North site. The control channel would bounce between correct and a 12GHz frequency, and the primary/secondary channels also showed incorrect.
Image

**Application Log**
Will send processing report to Dennis on Messenger unless something else is needed

**Desktop (optional - complete the following information):**
 - OS: Win11 64bit
 - CPU Cores: 8c/16t Ryzen 7 3700X
 - RAM: 32GB DDR4

**Additional context**
While typing this post, I reset the channel and all appeared to go back to normal, and has not glitched out since.

utafrali:
The decoder was reusing frequency data from the old control channel when switching to a new one. Cleared the map on channel switch and simplified the storage logic to always update instead of trying to avoid duplicates. Fixes #2298

### DIFF
--- a/src/main/java/io/github/dsheirer/module/decode/p25/P25TrafficChannelManager.java
+++ b/src/main/java/io/github/dsheirer/module/decode/p25/P25TrafficChannelManager.java
@@ -199,6 +199,7 @@ public class P25TrafficChannelManager extends TrafficChannelManager implements I
 
             mTS1ChannelGrantEventMap.clear();
             mTS2ChannelGrantEventMap.clear();
+            mFrequencyBandMap.clear();
 
             //Remove the control channel from the previous frequency
             mAllocatedTrafficChannelMap.remove(previous);
--- a/src/main/java/io/github/dsheirer/module/decode/p25/phase1/P25P1MessageProcessor.java
+++ b/src/main/java/io/github/dsheirer/module/decode/p25/phase1/P25P1MessageProcessor.java
@@ -330,14 +330,7 @@ public class P25P1MessageProcessor implements Listener<IMessage>
         if(message instanceof IFrequencyBand)
         {
             IFrequencyBand bandIdentifier = (IFrequencyBand)message;
-
-            //Only store the frequency band if it's new so we don't hold on to more than one instance of the
-            //frequency band message.  Otherwise, we'll hold on to several instances of each message as they get
-            //injected into other messages with channel information.
-            if(!mFrequencyBandMap.containsKey(bandIdentifier.getIdentifier()))
-            {
-                mFrequencyBandMap.put(bandIdentifier.getIdentifier(), bandIdentifier);
-            }
+            mFrequencyBandMap.put(bandIdentifier.getIdentifier(), bandIdentifier);
         }
     }
 


bazineta:
Curious as to the rationale for the `mFrequencyBandMap.clear()`. I wasn't aware that there was anything in there that was specific to the control channel; I thought that no matter what control channel you got that information from, it'd be the same information.

utafrali:
Good question. The `clear()` is a defensive measure for edge cases where the control channel switches to a different site or system. While in normal operation the frequency band identifiers should be consistent across control channels within the same system, during a control channel handoff to a different site the band map could carry stale entries that map to incorrect frequencies. Clearing it ensures we rebuild from the new control channel's broadcast data, avoiding any potential frequency lookup mismatches.

That said, if in practice the band map is always system-wide and never site-specific, the clear could be overly cautious. Happy to hear your thoughts on whether you've seen cases where the data differs across control channels.

bazineta:
> Happy to hear your thoughts on whether you've seen cases where the data differs across control channels.

In at least my local sites, (Phase 1 / Phase 2 hybrid simulcast) there will be 3 control channels, a low, middle, and high, and nearly 100% of the time they run on the middle channel; I've never seen a switch to either of the other two, but I presume it must happen for maintenance, etc., from time to time.

However, different control channel, but same site, so the data would be the same, i.e., it's not going to switch to a completely different site, so I propose that as a defensive measure, it's an extreme one.

TRENT310:
@bazineta while this may be the case on a small regional system, on a large system especially those operating on VHF/UHF bands or a mixture of sites in various frequency bands where there is inconsistent step/offset (as a matter of spectrum regulation, or lack thereof) there can be different band plans advertised on different sites control channels even of the same system, and likewise irrelevant band plans are not always advertised on all sites system-wide as part of the IDEN_UP and IDEN_UP_VU opcodes they are only broadcast on the sites that use them and the neighbouring sites that advertise them via their ADJ_STS_BCST.

Also some systems are designed with preferred/primary control channels which they will utilize consistently unless there is a fault or inbound interference condition, while other manufacturers and their implementation strategies allow all channels at a site to be a standby control channel, are advertised via the SCCB opcodes, and some even intentionally rotate the control channels periodically as part of a "wear leveling" strategy.

This "defensive measure" to handle the bigger picture issue where SDRTrunk does not currently offer NAC or WACN/System ID/RFSS/Site ID (NET_STS_BCST or RFSS_STS_BCST opcodes) validation in the playlist/channels and just merely utilize a list of frequencies which it steps through to find any control channel matching the configured protocol, regardless of the system identification. The System/Site/Name fields in each playlist are merely for user labeling purposes, and are not used to qualify the system being received is the correct one. While this is "convenient" for the casual listener who may not be familiar with the system parameters and is the quickest way to "just start hearing things" on the flipside for advanced use cases where the software is used for logging activity on a particular site or a particular talkgroup activity this flexibility can cause the software to get distracted and begin to follow the wrong site or system, missing out on traffic of interest. In the case of large, dense, or spectrally-congested system environments, frequencies are reused throughout, and well-placed receiver sites may receive multiple control or traffic channels including those of an unrelated site or system. Or, the SDRtrunk instance may be used while in motion while naturally travelling in and out of coverage just like a regular subscriber unit radio.

utafrali:
That makes sense - larger systems with cross-site band plans definitely need the clear(). System-level validation would be the proper fix, but this covers it in the meantime.

DSheirer:
Thanks for the PR!

PR #2414, opened 2026-05-18
#7 Tuner Auto-PPM Redesign With Channel Auto-PPM

DSheirer:
#7 WIP - redesigns the tuner auto-ppm and adds tuner and channel frequency correction managers


msullivan1993:
Dennis, is the NXDN Decoder branch worth playing with yet or still very much a WIP? I know of some NXDN in my area to test with. 

On the PPM corrections and whatnot, could we also get a way to affix tuners to a preset frequency on reboot and/or shutdown? I have one tuner that, when it comes back up, has a NOAA weather radio birdie right on one of my railroad frequencies and I have to manually change it every time.. lol

PR #2423, opened 2026-06-11
Extend supported DCS codes to 104 commonly supported in modern radios

pauldonahue:
I extended the list of supported DCS codes from original 83 found in ETSI TS 103 236 V1.1.1 to the commonly used 104 codes supported in most modern-day radios by calculating the Golay values, applying the offset and storing the values bit reversed in decimal format.

The formula copilot and I used for calculating the values was derived from: https://github.com/armel/uv-k5-firmware-custom/blob/main/dcs.c which is an open source firmware used on very popular cheap HAM radios.

Decoding has been tested against all new normal codes utilizing a combination of Quansheng, Baofeng, Radtel and Vertex Standard radios.

Paul Donahue - paul.donahue@gmail.com


pauldonahue:
Closing PR - only intended to include one commit not all.

PR #2427, opened 2026-07-02
Add JVM autoprovisioning in Gradle

yuzawa-san:
gradle has the ability to [auto-provision jvm's](https://docs.gradle.org/current/userguide/toolchains.html), so i put together this PR to do this. i use a different jdk (without javafx) by default and found it slightly annoying having to switch back and forth. it appears if you have the proper matching jdk installed it will not provision one, so your local environment and CI will be just as they were.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -66,11 +66,6 @@ java {
     }
 }
 
-sourceSets {
-    main.java.srcDirs 'src/main'
-    test.java.srcDirs 'src/test'
-}
-
 test {
     useJUnitPlatform()
 }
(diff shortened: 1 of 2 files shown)


src/main/java/io/github/dsheirer/audio/broadcast/BroadcastModel.java
DSheirer, line 1: Could you please remove from this PR the changes to BroadcastModel and the repackaging of the OpenMHzEditor?

yuzawa-san: @DSheirer done. however i would like to propose the changes still, but in a new PR https://github.com/DSheirer/sdrtrunk/pull/2434

DSheirer:
Thanks for the PR!

Issue #2428, opened 2026-07-02
Polyphase channelizer allocates per-block interim buffer + accumulator on every block


jstockdale:
# Polyphase channelizer allocates a per-block interim buffer + accumulator on every block (~0.4–1.6 GB/s of garbage)

**sdrtrunk Version**
master-branch

**Type**
Performance / enhancement

**Description**

`ComplexPolyphaseChannelizerM2.process()` allocates three `float[]` arrays on every invocation:

* `inlineInterimOutput` — length `subChannelCount × tapsPerChannel` (the largest)
* `filterAccumulator` — length `subChannelCount`
* `processed` — length `subChannelCount` (escapes to the IFFT thread)

`process()` is called once per sample block, and the block rate is fixed at `2 × DEFAULT_MINIMUM_CHANNEL_BANDWIDTH = 50,000 blocks/second` independent of sample rate (more channels ⇒ larger block ⇒ constant block rate). The allocation therefore scales with channel count:

* 96 channels (2.4 MHz): ~8.4 KB/call × 50k ≈ **403 MB/s**
* 400 channels (10 MHz): ~35 KB/call × 50k ≈ **1.64 GB/s**

These are short-lived allocations, so the cost is young-GC frequency and memory bandwidth rather than heap growth, but at multi-MHz sample rates it is a meaningful, continuous churn from a single hot method.

**Opportunity**

The `inlineInterimOutput` array is only used to hold `samples × filter` before the tap-sum; the multiply can be fused directly into the per-sub-channel accumulation, removing that array (and a redundant pass over it) entirely. The `filterAccumulator` is pure scratch that never escapes `process()` and can be reused across calls. Only `processed` must remain a per-call allocation because it is consumed on the IFFT thread. `process()` runs on a single thread (the `PolyphaseChannelManager` buffer dispatcher), so reusing the accumulator is safe.

This reduces per-call `float[]` payload ~11× and allocation events from 3 to 1, with bit-for-bit identical output (the tap products are summed in the same order; no FMA contraction occurs).

Happy to open a PR — I have the change and a JUnit equivalence test ready.

msullivan1993:
Would this fix the issue I have been running into in #2418 ? If so, I'd gladly love to test it in my build and see if it fixes things.

jstockdale:
[...]

Issue #560, opened 2019-05-25
Java Exception Out of Memory


radito:
I just got this error today. This thing happens when it started decoding p25 calls. And suddenly the app is crashing.
```
My system
OS: Kali GNU/Linux Rolling x86_64
Kernel: Linux dream 5.0.1 #1 SMP Tue Mar 12 10:12:37 WIB 2019 x86_64 GNU/Linux
Java: 
java version "11.0.2" 2019-01-15 LTS
Java(TM) SE Runtime Environment 18.9 (build 11.0.2+9-LTS)
Java HotSpot(TM) 64-Bit Server VM 18.9 (build 11.0.2+9-LTS, mixed mode)

```
```
07:37:31.312 INFO  i.g.d.m.d.p.a.P25AudioModule - Loading JMBE library from [/home/radito/jmbe/build/libs/jmbe-0.3.3.jar]  [86MB/124MB 69%]
07:37:31.367 INFO  i.g.d.m.d.p.a.P25AudioModule - JMBE audio conversion library successfully loaded - P25 audio will be available  [86MB/124MB 70%]
07:37:31.440 DEBUG i.g.d.d.f.c.ComplexPolyphaseChannelizerM2 - Sample Rate [2880000.0] providing [114] channels at [25263.157894736843] Hz each  [63MB/124MB 50%]
07:38:10.332 INFO  i.g.d.s.t.FrequencyErrorCorrectionManager - Auto-Correcting Tuner PPM to [-9.982047795254886]  [1GB/1GB 97%]
07:38:30.048 ERROR i.g.d.d.f.c.PolyphaseChannelManager - Error  [1GB/1GB 99%]
java.lang.OutOfMemoryError: Java heap space
        at io.github.dsheirer.sample.buffer.ReusableChannelResultsBuffer.getEmptyBuffer(ReusableChannelResultsBuffer.java:69)
        at io.github.dsheirer.dsp.filter.channelizer.ComplexPolyphaseChannelizerM2.process(ComplexPolyphaseChannelizerM2.java:357)
        at io.github.dsheirer.dsp.filter.channelizer.ComplexPolyphaseChannelizerM2.receive(ComplexPolyphaseChannelizerM2.java:218)
        at io.github.dsheirer.dsp.filter.channelizer.PolyphaseChannelManager$BufferSourceEventMonitor.receive(PolyphaseChannelManager.java:596)
[...]

Issue #2412, opened 2026-05-15
P25 Conventional missing calls


someengineer1:
Using various builds after 0.6.1 I've noticed a significant number of P25P1 calls (conventional system, no CC) getting missed.  Can see them on the waterfall, sometimes they even show in the events list, but are missing info and never get sent to Rdio Scanner.

I know at one point after 0.6.1 there were some issues with P25, have those all been resolved or are there still some lingering problems?  One such missed call came through as "encrypted unit to unit" even though we have no encryption and this frequency does not have any unit to unit ability, only broadcast.

msullivan1993:
It’s hilarious. I just asked Dennis this earlier today and was going to make a report tonight when I got home (which I just did) There’s no real commonality other than P25 conventional channels. But the last few days I’ve had channels randomly just stop decoding and capturing calls. 

Might have to vibecode some logging to see what’s going on unless the built in log will capture it. I am running a high capacity server so I hope that’s not the reason why.

someengineer1:
> I am running a high capacity server so I hope that’s not the reason why.

I've really been noticing this ever since I started testing your CTCSS and the other guy's too.  Prior to that I hadn't used SDRT in a while (and had only used it on Windows).  So I'm not sure at what point it changed but I've noticed it with all the testing I've done in the past few months on these builds.

Really my only reason to consider going back to SDRT is the better decoding on P25 (especially on my LSM system) so if that's not working right, everything else is somewhat moot.

My "server" is only monitoring a few things and is running less than 25% CPU with plenty of available memory.  I have 4 SDRs but that's always been the case.  I can't seem to determine any rhyme or reason but I'm wondering if it has to do with the extraneous ignored stuff that gets logged (some nearby system that logs a message but never locks on or decodes, but possibly ties up the SDR or something).

Last night I even had a couple out of order calls come in.  Just to be safe I disabled waterfall, minimized the app, signed out of nomachine, etc.  Does not seem to be resource related.

msullivan1993:
> > I am running a high capacity server so I hope that’s not the reason why.
> 
> I've really been noticing this ever since I started testing your CTCSS and the other guy's too. Prior to that I hadn't used SDRT in a while (and had only used it on Windows). So I'm not sure at what point it changed but I've noticed it with all the testing I've done in the past few months on these builds.
> 
> Really my only reason to consider going back to SDRT is the better decoding on P25 (especially on my LSM system) so if that's not working right, everything else is somewhat moot.
[...]

PR #2429, opened 2026-07-04
Reduce polyphase channelizer per-block allocation via fused multiply-accumulate

jstockdale:
## Summary

`ComplexPolyphaseChannelizerM2.process()` allocates three `float[]` on every call — an interim product buffer, the sub-channel accumulator, and the results array — and runs at a fixed `2 × DEFAULT_MINIMUM_CHANNEL_BANDWIDTH = 50,000 calls/second`, independent of sample rate. That's ~400 MB/s of short-lived garbage at 96 channels (2.4 MHz), rising to ~1.6 GB/s at 400 channels (10 MHz).

This fuses the multiply into the accumulate loop, which **removes the interim array entirely** (the largest allocation, plus a redundant pass over it) and **reuses the accumulator** across calls. The results array (`processed`) stays a per-call allocation because it escapes to the IFFT thread. Net: per-call `float[]` payload drops ~11×, allocation events 3 → 1. These are Eden-short-lived allocations, so the win is young-GC frequency and memory bandwidth, not heap footprint.

## Why it's safe

* **Bit-for-bit identical.** `mInlineSamples` and `mInlineFilter` are both exactly `subChannelCount × tapsPerChannel` (see `init()`), so the fused loop touches every element once and each sub-channel sums its tap products in the same order as before — the products are unchanged, and the JVM won't contract `fmul`/`fadd` into an FMA without an explicit `Math.fma()`. The added `ComplexPolyphaseChannelizerM2Test` asserts identity via `Float.floatToRawIntBits` across 8 channel counts (2–400) and 4 input distributions (incl. zeros, wide dynamic range, `-0.0`/subnormals): 32,000 comparisons, zero differences.
* **Reused accumulator is thread-safe.** `process()` runs only on `PolyphaseChannelManager`'s single-threaded buffer dispatcher; the IFFT runs on a separate thread and only touches the fresh `processed` arrays. The accumulator is (re)sized in `init()` like the other rate-dependent buffers.
* **No buffer pooling reintroduced.** It reuses only the thread-confined scratch accumulator — not the cross-thread `processed` array — so it avoids the reference-counting/OOM class of bug from the old `ReusableChannelResultsBuffer` (#560, #867).

## Testing

`./gradlew compileJava compileTestJava test` passes on JDK 25 (Liberica Full), including the new equivalence test. Output is bit-identical, so decode/audio behavior is unchanged. Happy to extend the test to drive the channelizer end-to-end if preferred.

Closes issue #2428

### DIFF
--- a/src/main/java/io/github/dsheirer/dsp/filter/channelizer/ComplexPolyphaseChannelizerM2.java
+++ b/src/main/java/io/github/dsheirer/dsp/filter/channelizer/ComplexPolyphaseChannelizerM2.java
@@ -1,6 +1,6 @@
 /*
  * *****************************************************************************
- * Copyright (C) 2014-2023 Dennis Sheirer
+ * Copyright (C) 2014-2026 Dennis Sheirer
  *
  * This program is free software: you can redistribute it and/or modify
  * it under the terms of the GNU General Public License as published by
@@ -24,6 +24,7 @@ import io.github.dsheirer.sample.complex.InterleavedComplexSamples;
 import io.github.dsheirer.util.Dispatcher;
 import java.text.DecimalFormat;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.List;
 import org.apache.commons.math3.util.FastMath;
 import org.jtransforms.fft.FloatFFT_1D;
@@ -46,10 +47,9 @@ import org.slf4j.LoggerFactory;
  *
  * Instead of using an array of channel filters as described in the Harris text, this filter and the sample buffer
  * are arranged as a contiguous array to maximize Java's ability to leverage native processor Single Instruction
- * Multiple Data (SIMD) intrinsics (since Java 8).  The filter process is broken into four steps:
+ * Multiple Data (SIMD) intrinsics (since Java 8).  The filter process is broken into three steps:
  *
- *   -Multiply the inline array of samples and filter coefficients
- *   -Accumulate the results for each sub-channel
+ *   -Fused multiply and accumulate of the inline sample/filter arrays into each I/Q sub-channel accumulator
  *   -Rearrange the sub-channel results to correctly order the sub-channels
  *   -Perform IFFT
  *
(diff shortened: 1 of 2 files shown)


yuzawa-san (commented):
oddly enough i was also looking into this a few days ago but you beat me to the pr. however it looks good. something i was going to add for bonus performance was to keep `inlineInterimOutput` and make it a field like you did with `filterAccumulator`. i found significant gains in doing a vectorized multiplication. 

i had a method like this:
```
private static void multiplyArrays(float[] a, float[] b, float[] c) {
      VectorSpecies<Float> species = FloatVector.SPECIES_PREFERRED;
      int speciesLength = species.length();
      int i = 0;
      int length = a.length;
      int upperBound = species.loopBound(length);
      while (i < upperBound) {
          FloatVector va = FloatVector.fromArray(species, a, i);
          FloatVector vb = FloatVector.fromArray(species, b, i);
          FloatVector vc = va.mul(vb);
          vc.intoArray(c, i);
          i += speciesLength;
      }
      while (i < length) {
          c[i] = a[i] * b[i];
          i++;
      }
    }
``` 

which i called using `multiplyArrays(inlineSamples, inlineFilter, inlineInterimOutput);`

there may be a way to do in one vectorized shot, but the addition is the cheaper part so i did not care to optimize it that much further. but maybe there is a way. i am new to the FloatVector world.

i was getting around to test this but ran into [an existing decoding issue](https://github.com/DSheirer/sdrtrunk/issues/2412#issuecomment-4883645408) on the master branch.

yuzawa-san (approved):
looks like the process to add a calibrated vectorized operation is more involved. lgtm as is.

DSheirer:
Awesome PR!  Thanks!

msullivan1993:
I finally got around to updating my build with the latest master (and this PR included) and I think these changes solved my issue with memory runaway. So far I have been able to bump 5 of my 6 RFOne's back to 10MHz with zero runaway (6th one is 2.5MHz because extra VHF is not needed currently, that one is just for railroad) Allocated memory is actually staying around 3GB without issue. Given my channel load, I consider that a win.

image

msullivan1993:
Update.. this fix has been huge for resource consumption still, at least RAM wise. Both my public safety ingests sit at less than 3-4GB allocated, and all my railroad feeds (which are anywhere from 2-20 NBFM channels) stay under 1GB allocated. I can't wait to stress-test my public safety machine with 8 10MHz tuners on it soon.

I will note, on one of my public safety ingests, I forgot to start the watchdog on it by accident (watches for excessive RAM allocation/usage and auto-restarts SDRTrunk at 05:30 every morning) and after 2-3 days, allocated did go up to 7GB (-Xmx8G) and was causing occasional sample drops, but that was from extended runtime. I don't know if something got stuck due to running for 2-3 days straight, but after restart it went back to normal. I should've done a dump before restarting to see what it was sticking around, but forgot.. just making a note of it for future reference.

@jstockdale you are the man.

PR #2434, opened 2026-07-09
Fix eclipse build

yuzawa-san:
i found these when getting my environment (eclipse on mac) set up for https://github.com/DSheirer/sdrtrunk/pull/2427
moved these changes upon request to this pr.
~additionally, i updated the gitignore to use the eclipse template.~ nevermind, i moved the rules to my global gitignore instead of this project's one.


DSheirer:
This PR contains only white-space and class renaming changes.  What is the issue that these changes are addressing?

yuzawa-san:
@DSheirer this is what the latest master returns in the most recent eclipse IDE version. i'm not sure why the gradle build seems to be more lenient, but the package and lambda usage is objectively incorrect and this fixes it.

image
image


image
image

DSheirer:
Thanks for the reply.  I agree, there is an error in the package declaration in the OpenMHzEditor class.

Can you please change your PR.  Instead of renaming the class path to match the package, can you correct the package declaration to match the class location:

```
package io.github.dsheirer.audio.broadcast.openmhz;
```

yuzawa-san:
my original rationale for the move was to keep consistency with where all of the other editors were located .

i have rebased and fixed the package references in 590f80ae

DSheirer:
Thanks for the PR!

PR #2436, opened 2026-07-10
Add Alias List column to the playlist Channels tab

TimmySheff:
Reference Feature Request #2216

This is a resubmission with a null check added.....


DSheirer:
Thanks for the PR!

PR #2437, opened 2026-07-10
Write aliases to CSV file

TimmySheff:
Creates the subdirectory "Aliases" and writes new (per session) radio aliases to a comma delimited file (.CSV).

This updates my prior PR to include NXDN aliases.


TimmySheff (commented):
Write new (per session) aliases to CSV files located in the "Aliases" subdirectory

DSheirer:
This PR adds File IO directly onto the demodulation thread and has the potential to weigh-down demodulation processing and cause unintended side-effects when running on a system that has a slow IO sub-system.  This could be problematic when running a large number of demod channels.

Consider adding a scheduled thread executor that fires periodically (1/minute?) and writes out any accumulated talker alias changes/additions.  That way the IO loading is predictable and most importantly, it runs on a separate thread.

TimmySheff:
No problem.  I will close and perhaps  ryanbytes will resubmit his PR (#2440
) since it sounds as if he has already tackled this.

PR #2411, opened 2026-05-14
Dark mode

AkumasCoffin:
brave_EMrZaRTA2O
brave_2RchRHr1d2
brave_hfO7VbUezH

PR #2440, opened 2026-07-11
Add safe opt-in talker alias CSV export

ryanbytes:
Provides a safer alternative implementation for the CSV-export portion of #2437 while remaining compatible with the alias-list column in #2436.

- Disabled by default and controlled in Application preferences.
- Writes per-system files under the application `Aliases` directory.
- Escapes CSV fields and sanitizes file names.
- Moves file I/O off decoder threads and serializes writers sharing a file.
- Logs directory and write failures instead of silently ignoring them.
- Preserves existing DMR, NXDN, and P25 alias processing paths.

### DIFF
--- a/src/main/java/io/github/dsheirer/identifier/alias/TalkerAliasManager.java
+++ b/src/main/java/io/github/dsheirer/identifier/alias/TalkerAliasManager.java
@@ -24,11 +24,21 @@ import io.github.dsheirer.identifier.IdentifierCollection;
 import io.github.dsheirer.identifier.MutableIdentifierCollection;
 import io.github.dsheirer.identifier.Role;
 import io.github.dsheirer.identifier.radio.RadioIdentifier;
+import io.github.dsheirer.preference.application.ApplicationPreference;
+import io.github.dsheirer.util.ThreadPool;
+import java.io.IOException;
+import java.nio.charset.StandardCharsets;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.StandardOpenOption;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
 import java.util.concurrent.ConcurrentHashMap;
+import org.slf4j.Logger;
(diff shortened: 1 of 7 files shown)


ryanbytes:
Closing at the project owner’s request; this change will remain in the downstream fork.

ryanbytes:
Reopening this in response to the discussion on #2437. This implementation adds an opt-in per-session CSV export for observed talker aliases, with coverage in `TalkerAliasManagerTest`.

src/main/java/io/github/dsheirer/identifier/alias/TalkerAliasManager.java
```diff
@@ -57,10 +81,57 @@ public void update(RadioIdentifier identifier, TalkerAliasIdentifier alias)
...
         }
     }
 
+    void appendCsv(int radio, String alias)
+    {
+        synchronized(EXPORT_LOCKS.computeIfAbsent(mExportFile, ignored -> new Object()))
+        {
+            try
+            {
+                Files.createDirectories(mExportFile.getParent());
```
DSheirer, line 151: Could you adjust this slightly to separate and move the file creation into the TalkerAliasManager constructor and introduce a non-static boolean flag indicating that the file was created and alias export is enabled.  Then, this method could check the state of the boolean and proceed or ignore.

If there's an issue with the IO system that prevents the file from being created, this method will log an IO error in the application log repeatedly for each new alias.

ryanbytes: Addressed in e0f2aef2. CSV initialization, including header creation, now occurs once in the manager constructor when export is enabled. A per-instance export flag prevents queued writes after initialization or write failures; a write failure disables further export attempts. Added coverage for constructor initialization and disabled export behavior.

DSheirer:
@ryanbytes apologies, I merged #2411 first, causing an conflict in ApplicationPreference with this PR.  I don't have write privilege to your repo, otherwise I'd fix it.

Can you please resolve the conflict?

